PR #1, opened 2016-11-16
Check whether Kite daemon is running/installed/authenticated/etc

alexflint:
This pr checks that Kite is installed, running, reachable, authenticated, and enabled in the current directory. If any of these checks fail then an appropriate noficiation is displayed with a button that lets the user fix the problem. It relies on the kite-installer package to perform the actual checks, and to do the installation / launching / etc.


lib/events.js
```diff
@@ -86,6 +88,10 @@ function httpRoundTrip(endpoint, obj) {
   };
 
   var req = http.request(options);
+  req.on('error', () => {
```
dhung09: may be useful to pass the error to mixpanel as well
```javascript
req.on('error', (err) => {
  // do something with `err`
});
```

lib/ready.js
```diff
@@ -0,0 +1,295 @@
...
+  },
+
+  // ensure checks that Kite is installed, running, reachable, authenticated,
+  // and enabled in the current directory. If any of these checks fail then an
+  // appropriate noficiation is displayed with a button that lets the user fix
+  // the problem.
+  ensure: function(notifyOnReady) {
+    var curpath = this.currentPath();
+    StateController.handleState(curpath).then((state) => {
+      console.log("state:", state);
```
dhung09: can probably remove this for prod

lib/ready.js
```diff
@@ -0,0 +1,295 @@
...
+        case StateController.STATES.AUTHENTICATED:
+          if (curpath !== null) {
+            this.warnNotWhitelisted(curpath);
+          } else if (notifyOnReady !== undefined) {
+            this.notifyReady();
+          }
+          break;
+        case StateController.STATES.WHITELISTED:
+          metrics.track("kite is ready");
+          if (notifyOnReady !== undefined) {
```
dhung09: seems like `notifyOnReady` is assumed to be a boolean and is optional? if so, you can change the signature to
```javascript
ensure: function(notifyOnReady=false) {
  // ...  
}
```
and then this check can just be written as
```javascript
if (notifyOnReady) {
  this.notifyReady();
}
```

lib/ready.js
```diff
@@ -0,0 +1,295 @@
...
+            notification.dismiss();
+            this.install();
+          }
+        }]
+      });
+      notification.onDidDismiss(() => {
+        metrics.track("download-and-install error dismissed");
+      });
+    });
+    // TODO: on failure display a notification with an option to retry
```
dhung09: stale comment?

lib/ready.js
```diff
@@ -0,0 +1,295 @@
...
+      }]
+    });
+    notification.onDidDismiss(() => {
+      metrics.track("not-running warning dismissed");
+    });
+  },
+
+  launch: function() {
+    metrics.track("launch started");
+    StateController.runKite().then(() => {
```
dhung09: we can use `runKiteAndWait` now and remove the `setTimeout` below

lib/ready.js
```diff
@@ -0,0 +1,295 @@
...
+          }
+        }]
+      });
+      notification.onDidDismiss(() => {
+        metrics.track("launch error dismissed");
+      });
+    });
+  },
+
+  warnNotReachable: function() {
```
dhung09: this may be an extremely rare edge case, but i wonder how many times this warning will show up during normal use? i.e. kite is not hanging, it's just taking a little bit of time for the http server to get ready and so the user mistakenly thinks something is wrong

lib/ready.js
```diff
@@ -0,0 +1,295 @@
...
+    notification.onDidDismiss(() => {
+      metrics.track("not-authenticated warning dismissed");
+    });
+  },
+
+  authenticate: function() {
+    metrics.track("authentication started");
+
+    // TODO: show some kind of login UI
+    var email = "test@kite.com";
```
dhung09: wonder if there are any built in atom api's we can use with low effort

lib/metrics.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+  return id;
+}
+
+// Send an event to mixpanel
+function track(eventName, properties) {
+  if (properties === undefined) {
+    console.log("event: " + eventName);
+  } else {
+    console.log("event: " + eventName, properties);
```
dhung09:
```javascript
console.log("event :${ eventName }", properties);
```

lib/metrics.js
```diff
@@ -0,0 +1,53 @@
...
+  var id = localconfig.get('distinctID');
+  if (id === undefined) {
+    id = crypto.randomBytes(32).toString('hex');
+    localconfig.set('distinctID', id);
+  }
+  return id;
+}
+
+// Send an event to mixpanel
+function track(eventName, properties) {
```
dhung09: i assume before we release this to general public, we'll have some sort of consent/opt-in?

lib/metrics.js
```diff
@@ -0,0 +1,53 @@
...
+    console.log("event: " + eventName, properties);
+  }
+  eventData = {
+      distinct_id: distinctID(),
+      atom_version: atom.getVersion(),
+      kite_plugin_version: kitePkg.version,
+      os: OS_VERSION,
+  };
+  if (properties !== undefined) {
+    for (var key in properties) {
```
dhung09: slightly more concise way to write this
```javascript
for (var key in properties || {}) {
  eventData[key] = properties[key];
}
```

lib/ready.js
alexflint: ah so much I still don't know about es6 :)

lib/ready.js
alexflint: was asking about this on the atom slack channel but nobody had any good suggestions other than building it ourselves, unfortunately, so I'll probably just do that now

lib/ready.js
alexflint: Yeah that's a good point. I have seen that exact situation a few times locally. I think I will remove this warning for now but leave the code there, and then we can put it back in if we see real reachability problems during testing.

lib/metrics.js
alexflint: Yes let's look into that next week.

PR #2, opened 2016-11-19
check whether the current dir is whitelisted when sending events to kited

alexflint:
This pr changes the kite plugin to check whether the current dir is whitelisted by checking the HTTP response code from the `/clientapi/editor/event` endpoint.

It also makes the following more minor changes:
- clicking a button in a notification does not trigger the one hour lockout for that notification.
- do not show launch error if the error was that kite is already running, and similarly for other errors
- bump up completions priority so that it always supercedes autocomplete-python
- add a "Reset password" button to the login UI

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -41,8 +41,8 @@ function reset() {
   PENDING_EVENTS = [];
 }
 
-// minimum interval in milliseconds between sending "could not connect..." events to mixpanel
-const CONNECT_ERROR_LOCKOUT = 15 * 60 * 1000;
+// minimum interval in seconds between sending "could not connect..." events to mixpanel
+const CONNECT_ERROR_LOCKOUT = 15 * 60;
 
 // last time we sent a "could not connect..." event to mixpanel
 var lastConnectError = null;
@@ -67,7 +67,7 @@ function mergeEvents() {
     }
   }, (err) => {
     // on connection error send a metric
-    if (lastConnectError === null || utils.timeSince(lastConnectError) >= CONNECT_ERROR_LOCKOUT) {
+    if (lastConnectError === null || utils.secondsSince(lastConnectError) >= CONNECT_ERROR_LOCKOUT) {
       lastConnectError = new Date();
       metrics.track("could not connect to event endpoint", err);
     }
--- a/lib/events.js
+++ b/lib/events.js
@@ -137,6 +137,18 @@ function onEdit(editor) {
 function onSelection(editor) {
   send(buildEvent(editor, "selection"));
 }
+function onActivate() {
+  // observeTextEditors takes a callback that fires whenever a new
+  // editor window is created. We use this to call "observeEditor",
+  // which registers edit/selection based callbacks.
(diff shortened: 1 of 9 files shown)


lib/utils.js
```diff
@@ -10,6 +10,13 @@ function pointToOffset(text, point) {
   return total;
 }
 
+// get time in milliseconds since the date
+function timeSince(when) {
```
dhung09: for clairty, maybe name this to `millisecondsSince`?

lib/events.js
```diff
@@ -54,7 +60,18 @@ function mergeEvents() {
   if (DEBUG) {
     console.log(event.action, event.filename, event.selections[0]);
   }
-  httpRoundTrip('/clientapi/editor/event', event)
+  httpRoundTrip('/clientapi/editor/event', event).then((resp) => {
+    if (resp.statusCode == 423) {
```
dhung09: `resp.statusCode === 423` in general its better practice to use `===` because it does type checking i.e. `423 == '423'` would equal true while `423 === '423'` is false

lib/ready.js
```diff
@@ -92,6 +81,38 @@ var Ready = {
     this.ensure(true);
   },
 
+  // ensureIfPython  ensure only if the currently active file extension matches that provided
+  ensureIfPython: function() {
```
dhung09: i think another way to do this is in `package.json`, you can define `activationHooks` i.e. in `autocomplete-python`: https://github.com/autocomplete-python/autocomplete-python/blob/master/package.json#L6-L10

we have this turned off in our fork because if it were on, the install flow wouldn't trigger automatically once it's first installed

lib/ready.js
```diff
@@ -132,6 +158,15 @@ var Ready = {
       this.launch();
     }, (err) => {
       metrics.track("download-and-install failed", err);
+
+      // if install failed because kite is already installed then ignore
+      var curState = err.data || 0;
+      if (curState >= StateController.STATES.INSTALLED) {
+        console.log("download-and-install failed because kite is already installed (state { curState })");
```
dhung09: if you're trying to templatize `curState`:
```javascript
console.log(`download-and-install failed because kite is already installed (state ${ curState })`);
```

lib/ready.js
```diff
@@ -300,6 +378,15 @@ var Ready = {
       this.ensure();
     }, (err) => {
       metrics.track("whitelisting failed", {dir: dirpath});
+
+      // if whitelist failed because dir is already whitelisted then ignore
+      var curState = err.data || 0;
+      if (curState >= StateController.STATES.WHITELISTED) {
+        console.log("whitelist failed because dir is already whitelisted (state { curState })");
```
dhung09:
```
console.log(`download-and-install failed because kite is already installed (state ${ curState })`);
```

lib/ready.js
alexflint: Ah good point. There is still some stuff in activate that needs to run no matter what, like the config observer, but I'd like to investigate this more in a future pr.

PR #3, opened 2016-11-22
Updated copy for notification text

adamsmith:
A few themes:

* Reframing from the Kite autocomplete daemon -> Kite Menubar app.  The thinking was 'daemon' sounds too technical, and we want to use common wording across e.g. the Menubar menu and these notifications.
* I tried to make the heading text shorter, and more like a title.
* My main concern was with the whitelist copy, which I made similar to what we currently have for that in the sidebar.

### DIFF
--- a/lib/ready.js
+++ b/lib/ready.js
@@ -95,7 +95,7 @@ var Ready = {
   warnNotSupported: function() {
     metrics.track("not-supported warning shown");
     atom.notifications.addError(
-      "OS not supported", {
+      "Kite doesn't support your OS", {
       description: "Sorry, the Kite autocomplete engine only supports macOS at the moment.",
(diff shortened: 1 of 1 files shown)


lib/ready.js
```diff
@@ -200,8 +200,8 @@ var Ready = {
   warnNotReachable: function() {
     metrics.track("not-reachable warning shown");
     atom.notifications.addError(
-      "The Kite autocomplete daemon is running but not reachable", {
-      description: "Try killing Kite from Activity Monitor.",
+      "The Kite background service is running but not reachable", {
```
alexflint: Hmmm but now it seems we have "Kite", "the Kite app", "Kite background service", and "Kite autocomplete engine".

PR #5, opened 2016-11-24
Code review

abe33:
🚧 **This is still a work in progress** 🚧 

I tried to make each commit as simple as I could, many can probably be cherry-picked but I'm not sure that there won't be any conflicts on some.

I'll probably also look at how refactor the code to get rid of all the boilerplate for notifications and such.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+node_modules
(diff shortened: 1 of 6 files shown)


abe33:
I was wondering, regarding the request made with the file content to kite, how do you handle file's encoding, because whether on Sublime Text or Atom we'll probably have to deal with all kind of encoding (ANSI VS Unicode is a pretty typical use case).

PR #6, opened 2016-11-28
Refactor ready.js

abe33:
This is an attempt to make the object thinner and to get rid of all the boilerplate.

### DIFF
--- a/lib/ready.js
+++ b/lib/ready.js
@@ -30,51 +30,40 @@ var Ready = {
   // appropriate noficiation is displayed with a button that lets the user fix
   // the problem.
   ensure (forceNotify=false) {
-    var curpath = this.currentPath();
-    StateController.handleState(curpath).then((state) => {
-      switch (state) {
-        case StateController.STATES.UNSUPPORTED:
-          if (this.shouldNotify(state) || forceNotify) {
-            this.warnNotSupported();
-          }
-          break;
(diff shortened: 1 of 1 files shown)


lib/ready.js
```diff
@@ -29,61 +29,50 @@ var Ready = {
...
-          if (forceNotify) {
-            this.notifyReady();
-          }
-          break;
-      }
-    }, (err) => {
-      metrics.track("handleState failed", err);
-    });
+  ensure (forceNotify=false) {
+    const curpath = this.currentPath();
```
alexflint: I feel like this change makes this function less easy to understand on net. I agree there is some code duplication in the original but isn't a little duplication okay if it's easy to understand?

abe33: True, ideally I would like `ensure` to only contains the action map and the invocation of the state controller. I think all the notifications related functions could be extracted from that file.

One of the big advantage of having a more functional approach is that you can more easily reason about the behaviour you want by combining smaller functions rather than modifying large chunk of code in several places. But it sure can be confusing at first, especially when coming from a more imperative background (it's my own experience so I know I had to study and practice a lot before being able to put myself in a functional mindset).

If you don't mind I would like to try pushing this idea a bit further, because I somehow stopped midway, trying to grasp the subtleties of the concrete notification methods.

PR #19, opened 2016-12-07
Update `distinctID` to user ID

### LINKED ISSUE #9: Get user ID from Kite for use in mixpanel events
Currently we use the editorUUID for all mixpanel events. However, once the user has Kite running and is logged in, we should use the following endpoint to get the user ID and then use that as the distinct_id in all future mixpanel events.

```
GET localhost:46624/clientapi/user

{"id":7,"name":"alex","email":"alex@kite.com","bio":"","email_verified":true,"is_internal":true,"unsubscribed":false}
```

(The editorUUID should still be added as a separate "editorUUID" field in the mixpanel event, but it should not be used as the distinct_id.)

dhung09:
Fixes #9 by adding a function `saveUserID` to the `ready.js` module which requests the current user's info from `kited`. `saveUserID` is called in any state after which the user should be authenticated.

Also updates `kite-installer` to `0.5.0` and updates the `StateController` calls accordingly

cc @alexflint @abe33


lib/ready.js
```diff
@@ -66,6 +68,35 @@ var Ready = {
     this.statusItem.onClick(this.ensure.bind(this));
   },
 
+  // saveUserID requests the current user's ID and saves it to localconfig.
+  // It does nothing if the current user is not authenticated or if there
+  // was an error fetching and parsing the user's info.
+  saveUserID: function() {
```
abe33: Shouldn't it be named with a name such as `fecthKiteUserID`?

lib/ready.js
```diff
@@ -66,6 +68,35 @@ var Ready = {
...
+    }, (resp) => {
+      if (resp.statusCode !== 200) {
+        return;
+      }
+      var raw = '';
+      resp.on('data', (chunk) => raw += chunk);
+      resp.on('end', () => {
+        try {
+          var data = JSON.parse(raw);
+          if (data.id !== undefined) {
```
abe33: `data.id != null` or `data.id != undefined` is generally fine because while `null == undefined` OTOH `null !== undefined`.

lib/metrics.js
```diff
@@ -34,6 +36,7 @@ function track(eventName, properties) {
   }
   var eventData = {
     distinct_id: distinctID(),
```
intrepidlemon: instead of manually sending the distinct_id with every event, you can also use https://mixpanel.com/help/reference/javascript-full-api-reference#mixpanel.identify 
when you first get user_id. I'm not sure but I think the advantage of using this is that mixpanel can retroactively label events that came from this session under the same user before they've logged in.

PR #20, opened 2016-12-08
Handle password less login

abe33:
So we start on that screen: 

kite-login

I made the submit button use a `btn-primary` class to follow usability best practice on primary action VS secondary actions.

Here's what the form look like when one of the field is invalid: 
kite-invalid-email
kite-invalid-password

It's clearly better than having the form disappear and a notification appear because you made a typo. 
I tried to use a `required` attribute on the fields but they initially appeared as invalid. I'll have to dig to find why it's like that in Atom.

And finally here's the version when we receive a code `9` for a user without a password: 

kite-no-password

The button has currently the same behaviour as the reset password button.

As for the wording, feel free to tell me what to use instead.


adamsmith:
Small nit: instead of "Send me a link to set my password" maybe we should use "Resend password setup email"?

lib/ready.js
```diff
@@ -312,16 +312,51 @@ var Ready = {
...
+            disposeUI()
+            break;
+          case 400:
+          case 401:
+            kiteUtils.handleResponseData(res, (raw) => {
+              const data = JSON.parse(raw)
+              console.log(data)
+              switch (data.code) {
+                // Password-less
+                case 6:
```
alexflint: Do these correspond to the codes from `StateManager` in kite-installer? If so, can you use those constants?

lib/elements/login.js
```diff
@@ -6,49 +6,28 @@ var LoginStep = class {
...
-    this.resetBtn.textContent = "Reset Password";
-    this.element.appendChild(this.resetBtn);
-
-    this.cancelBtn = document.createElement('button');
-    this.cancelBtn.classList.add('btn');
-    this.cancelBtn.classList.add('btn-large');
-    this.cancelBtn.classList.add('inline-block');
-    this.cancelBtn.textContent = "Cancel";
-    this.element.appendChild(this.cancelBtn);
+    this.element.innerHTML = `
```
alexflint: Oh this is so much better - I did not know about this : )

lib/ready.js
abe33: They're coming from https://github.com/kiteco/kiteco/blob/master/kite-go/community/errors.go

I can still use constants though, but shouldn't they come from the `AccountManager` then?

PR #22, opened 2016-12-09
Use modal panel for login form

abe33:
Only consider the last commit, I created this branch from the password-less one to prevent merge conflicts.

So the main changes are: 

- Using a modal panel instead of a top panel
- Reorganize the form to make it more compact
- Allow the `core:cancel` command to close the form (Typically bound to the <kbd>esc</kbd> key on OSX). It's basically just a binding to the cancel button.
- Put the initial focus in the email field

As for the look and feel, it now looks like this with the `one-dark` theme (Atom's default): 

ready_js_ ___github_kite

I think it'd look prettier with the kite logo at the top of the modal, what do you think?


adamsmith:
Awesome stuff!!!

Small tweak: "Sign in to Kite" -> "Sign into Kite"

Also, is it possible to make "Reset password" a link with small font-size right justified under the Password field?

abe33:
Here's how the new version looks: 

login_less_ ___github_kite

adamsmith:
The UI looks good to me.  @dhung09 maybe you can do the code review?

There's a meta discussion here around code reviews: do we need them for this repo, who on the Kite team should do most of the reviews.  Maybe we can discuss on our phone call soon?

dhung09:
re: reviews i think we should still do reviews for this repo's since it's public facing. ideally we'd have automated tests for another layer of verification that nothing is broken, but for now i'm happy to help out with code reviews where necessary

Issue #21, opened 2016-12-09
If user uninstalls Kite.app, don't show "Install Kite" notification


adamsmith:
We should set a flag the first time we see Kite.app installed.  If we later detect that Kite is NOT installed we shouldn't show the "Install Kite" notification since the user (most likely) just uninstalled Kite.

PR #26, opened 2017-01-02
Don't display notification if Kite was uninstalled by the user

abe33:
Ref #21 

Note that if the user has set the autocomplete-python engine to kite, even if the notification don’t show up, the engine choice form still kicks in.

### DIFF
--- a/lib/ready.js
+++ b/lib/ready.js
@@ -83,6 +83,13 @@ var Ready = {
     var curpath = this.currentPath();
     StateController.handleState(curpath).then((state) => {
       this.updateStatusItem(state);
+
+      // when state is greater than or equal to STATES.INSTALLED we're sure
+      // kite have been installed once so we'll store that for later use.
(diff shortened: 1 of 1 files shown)


lib/ready.js
```diff
@@ -83,6 +83,13 @@ var Ready = {
     var curpath = this.currentPath();
     StateController.handleState(curpath).then((state) => {
       this.updateStatusItem(state);
+
+      // when state is greater than or equal to STATES.INSTALLED we're sure
+      // kite have been installed once so we'll store that for later use.
+      if (state >= StateController.STATES.INSTALLED) {
+        localStorage.setItem('kite.wasInstalled', true)
```
alexflint: nit: missing semicolon at end of line

abe33: I think I'll merge the pr as it, then merge the eslint one and fix that using eslint. It'll be simpler that way.

PR #36, opened 2017-01-19
Allow to choose the root (homedir or /) or to pick a directory

abe33:
The notification now looks like this: 

ready_js_ ___github_kite

When choosing `browse` the user is presented with a native dialog to pick a directory to whitelist.


adamsmith:
From the UI perspective looks great!!  (Thanks for instrumenting with metrics, also!)

PR #38, opened 2017-01-25
Style tweaks

abmirayo:
@abe33 cc @adamsmith 

Some design tweaks as part of a design iteration:

- Added small padding to section contents
- Changed design of *Usages* section
- Changed design of the footer
- Small changes to sizes, paddings, and colors. 

Also added comments at https://kite.quip.com/acH6ATLLWrDE

image

### DIFF
--- a/lib/elements/kite-expand-function.js
+++ b/lib/elements/kite-expand-function.js
@@ -20,10 +20,8 @@ class KiteExpandFunction extends HTMLElement {
     </section>
 
     ${renderParameters(symbol)}
-    <section>
-      ${renderInvocations(symbol)}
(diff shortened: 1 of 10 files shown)


lib/elements/kite-expand-function.js
```diff
@@ -20,10 +20,8 @@ class KiteExpandFunction extends HTMLElement {
     </section>
 
     ${renderParameters(symbol)}
-    <section>
```
abe33: This wrapper section was there so that we can group the two sub-sections together in one column when using the block decoration version (which is rendered inline). I'll use a different wrapper so that the section styles don't apply to that wrapper

PR #39, opened 2017-01-30
Sidebar UI

abe33:


abmirayo:
I'll start leaving some design comments.

1. The delay between double click and showing the expanded UI is pretty high. I know we changed the delay for the mouse hover UI, but did we also change this one, or is this loading time?

2. When clicking "module", I get a warning icon, with the "forbidden" cursor and a 404 tooltip. The main issues are: warning icon is too dramatic for a 404, the tooltip only shows up if i move away the cursor and hover it again (most people won't do that) and the tooltip shows 404 instead of a legible user-facing error. For these cases, the ideal behavior would be:
  - User clicks on "module"
  - We get 404 error. Instead of showing a warning icon, we just gray out the link.
  - We immediately trigger a very lightweight tooltip saying "No results available for X"

demopr

abmirayo:
4. Can we move the back and forwards button area to where it was before, at the bottom left of the overlay, vertically aligned with the "Open in Web" area?

demopr

5. Once moved to the bottom left, I'd either:
- Remove the interactive breadcrumb 
- Or limit the breadcrumb to the previous node and the current one

abmirayo:
6. Weird mouse scrolling behavior where I can scroll to the right hidden screen
demopr

7. The expanded members area is very clever. Some comments:
  - In the "search box", add placeholder text, grayed out, with "Filter by name".
  - Change color of "Showing all of X members" to standard white/gray
  - Change color of every node link the list to "link blue" color.
demopr

abmirayo:
8. Make top member links be "link blue" color
image

abmirayo:
9. I don't know if this PR has the completions in scope. If it does, can we limit the length of the description of each token and limit the width of the completions box as well?

image

abmirayo:
10. In general, once we have examples and one-liner usages, I'd put those at the top, before the description, so that we prioritize practical content but the user is still able to scroll down and get all of the description.

abmirayo:
@abe33 I have to say, this is all very, very impressive. I have mixed feelings about the breadcrumb and the members view, since I initially envisioned the overlay as much more lightweight and with minimal navigation, but I'd leave it to see what user response is like and I hope I'm proven wrong. 

Thank you for taking the creative license to make all these tweaks, the members view, the "Expand" button on the hover UI. This is all very fresh and awesome. I'm loving it. I just left a few comments above.

abe33:
@abmirayo 

1. I think it's the loading time, we're actually doing several roundtrip to the server, one to check if the file is whitelisted, one to get the hover data and then retrieve the report. I think I can improve that by doing the whitelist check much sooner and only once.

2. Yeah, totally make sense. At first I just wanted something that works so that you can see what to improve.

4. Sure
5. Shouldn't be a problem. I don't know if you noticed, but the breadcrumb is scrollable and is actually scrolling to always display the current step.

6. Yeah, I still need to improve this part, actually I'm using a nice CSS trick with transform and data attributes to make the whole thing scroll on navigation, I should be able to prevent manual scrolling using and `overflow property`.

7. 👇 
  - No pb
  - No pb
  - I would say that this go against the current behaviour of Atom, where's list item are in grey and filter matches are in blue: data-loader_js_ ___github_kite
  BTW, I was thinking of using Atom's list styles for that section but dropped the idea to follow your initial design, what's your thoughts about that?

8. Only for this one of for all the section titles as well?

9. It was not in the scope, but this is something I can fix on master. On that note, maybe we should think of a getting a nicely formatted short description, as is actually this is mostly a big blob of text, and when using the html version it's even worse because autocomplete-python is meddling with autocomplete-plus styles.

10. Yep make sense.

BTW, did you try the other configurations? (sidebar and block display) A breadcrumb and navigation at the top make more sense in that context. Happily this is through two different elements so it can be adjusted by context to some extend.

abmirayo:
@abe33 

1. Whitelist lookup should be instant, so I don't know if that'd would help, but other requests do have to hit the servers. I tried using the network panel from Chromium to get network latency but I'm not getting anything on the requests the plugin is doing.

abe33:
@abmirayo 

Re 4: that creates a problem, the navigation isn't part of a step's content, while the open in web is. That means the button slides with the content while the navigation isn't. Not sure how to solve this.

abe33:
I played around with the current markup but using absolute positionning to deal with thez navigation at the bottom, and what I feared occured, below is what happens when the two items in the breadcrumb are too wide for the available space: 

sample_py_ ___github_kite

but as it uses absolute positionning we can't reliably make it fit the space on the left of the `open in web` button

PR #94, opened 2017-03-02
Small tweaks + branding

abmirayo:
@abe33 cc @adamsmith 

Some small tweaks to hover + branding

image


abe33:
That's good for me

PR #97, opened 2017-03-06
Feature 403 handling

abe33:


jlozano:
hey man!

Looking good thus far!

couple of minor things:
- [ ] instead of button saying `Ignore` in the whitelist modal can we change it to say `Settings`
- [ ] instead of putting to the blacklist when the user hits `Ignore` we should launch the user's browser pointed at `local.kite.com:46624/settings/permissions?filename={filename}&action=blacklist` this endpoint will be in place today, will ping you when it gets merged in

jlozano:
@abe33 one last thing, when you call `PUT /clientapi/permissions/blacklist` the file in the json body should be the path to the current file the user is editing.

PR #106, opened 2017-03-10
Status panel

abe33:


abmirayo:
@abe33 Love it. Can we separate "Kite Basic" from the logo just a tiny bit more?
image

abmirayo:
Can we replace the Kite Pro bit with the logo? (Instead of using text "Pro")

Assets here:
https://www.dropbox.com/s/vtet730h8j2tov1/icon-pro-dark.png?dl=0
https://www.dropbox.com/s/zjbnexsevy2lb6k/icon-pro-light.png?dl=0

abe33:
@abmirayo I'd rather use a svg version if you have that, because it'll be more consistent with how we display the kite logo currently.

PR #150, opened 2017-03-30
Feature active search

abe33:


abmirayo:
@abe33 I'm going to add some design comments in this thread

abmirayo:
- I think we should offset the `box-shadow` a little bit horizontally, so there's no box-shadow on the right, and have the `right` positioning be `1px`, since Atom's border around the editor is 1px but absolutely positioned elements ignore borders:

```box-shadow: -5px 0 10px rgba(0, 0, 0, 0.2);```
```right: 1px;```

(This is for top-right positioning, for left positioning it would be left: 1px and a positive 5px horizontal offset in the box-shadow) 

Result would be:
image

abmirayo:
- Can we make the padding-left of kite-active-search component-padding instead of double-padding?

abmirayo:
- Could we make the animation be only horizontal? Like this:

demo

abmirayo:
- Can we make the input have sans-serif font instead of monospace?
image

abmirayo:
@abe33 Something else I've noticed is that both in the expand panel and in active search, we have a problem with long summary sections, since the user doesn't see examples/parameters/links/usages because they're too far, and scroll isn't as intuitive (especially on macOS, where scrollbars are hidden if using an Apple mouse/trackpad).

The idea would be to have sticky section headers at the bottom that follow the user scroll and users can also click on them to scroll automatically. I've made a fiddle to illustrate it: https://jsfiddle.net/antoniobustamante/r8utgq4z/

This is probably a bigger change, and it affects both the expand panel and active search, so maybe we should include this in a separate PR? (Also the fiddle is very plain HTML, plain JS. I'm sure there's a more reasonable way of doing it inside of the architecture you've built of components)

abmirayo:
@abe33 Everything else is great. It's borderline impressive and works great. Thanks a lot, it's looking :1st_place_medal:

Issue #178, opened 2017-05-13
Status panel tweaks


jlozano:
@abe33 
- we added to more states for the status, `Kite is indexing your code` and `Kite is syncing your code`, they should both have the same blue icon that we show when the plugin is functioning normally. The status can be retrieved from `/clientapi/status`, see here for api: https://kite.quip.com/jJdMAx2b8OsT#AaQACANNDrD

- we added the information on user's referrals to the response from `/clientapi/plan` so that we can implement the "invite" calls to action, see here for api: https://kite.quip.com/PgOaAiA5GUUM.

full design doc: https://www.dropbox.com/s/0vlmqw9bws0sbiy/PluginsAudit_April2017-02.jpg?dl=0
api doc: https://kite.quip.com/rVcJA1bIqZny

jlozano:
@abe33 nice work!

jlozano:
@abe33 one minor thing, the link for the "Referral Status" to "Invite more people" should link to `localhost:466234/redirect/invite`. https://kite.quip.com/rVcJA1bIqZny#QcWACAfPRvB

abe33:
1. I believe the current url is correct, or did I miss something? https://github.com/kiteco/atom-plugin/blob/master/lib/elements/kite-status-panel.js#L229

jlozano:
@abe33 weird, will double check today, that link is correct, not sure what i saw yesterday.

jlozano:
@abe33 confirmed 1 is the correct url, sorry about that. confirmed 2 and 3 are fixed, great work on this!

PR #179, opened 2017-05-15
Feature extended status

abe33:


abe33:
Ref #178

jlozano:
Looking good! Some things I noticed:
- The link for `Account` when a user is pro should link to `GET localhost:46624/clientapi/desktoplogin?d=/settings/acccount`, I think it currently links to `kite.com/pro` ?
- It looks like we are not showing the referral status message: https://kite.quip.com/rVcJA1bIqZny#QcWACAfPRvB ?

Note: I added some debug plan states that can be manually set in the menu bar, this will be released tomorrow and can be used for testing.

abe33:
It should be fine for both bullet points

PR #180, opened 2017-05-16
Better unsupported message

abe33:
For windows it now displays a specific message when it's the architecture that mismatch: 
```
Sorry, the Kite engine only supports Windows7 and higher with a 64bit architecture. 
Your version is actually recognised as: {{arch}}
```

For Linux we're using the old version:
```
Sorry, the Kite engine only supports macOS and Windows at the moment.
```

For OSX we're now mentioning the minimum requirements: 
```
Sorry, the Kite engine only supports OSX 10.10 (Yosemite) and higher.
```


jlozano:
LGTM pending CI fixes!

Issue #176, opened 2017-05-13
Design pass


jlozano:
Some updates with design docs:
 - Expand panel: https://www.dropbox.com/s/3z6zkhlwsv07928/PluginsAudit_April2017-03.jpg?dl=0
    - Show parameters and types below signature (for functions)
    - Show Popular patterns below parameters (for functions)
    - Show Inferred **kwargs below popular patterns (for functions) check out https://kite.quip.com/RqgeAKbfeFax#CGEACAGOHVW for examples of inferred **kwargs, in particular the `kwarg_parameters` field.
 - Hover: https://www.dropbox.com/s/x4jssc8q50nn2ib/PluginsAudit_April2017-04.jpg?dl=0
    - Just show the type name and the kind, for functions do *not* show the arg spec.

jlozano:
looks great!

PR #181, opened 2017-05-17
Feature design pass

abe33:
Ref #176


jlozano:
@abe33 looking really good! 2 minor things:
- in the expand panel can we switch the ordering of the sections to:  signature, parameters with types, popular patterns, **kwargs, docs, examples, usages, definition, links?
- can we include the inferred types for the **kwargs section in the live call signature?

abe33:
The sections are reorganized!

Issue #177, opened 2017-05-13
Include **kwargs in live signatures


jlozano:
@abe33  the idea here is to allow the user to see the inferred **kwargs by clicking on that field in the call signature (see design doc).

https://www.dropbox.com/s/pmmwqyqw0pqgh44/LiveCallSignatures-01.jpg?dl=0

jlozano:
looks great!

PR #182, opened 2017-05-17
Display kwargs in signature panel

abe33:
Ref #177


jlozano:
@abe33 this is awesome! One minor thing: 
- can we include the inferred types for the **kwargs section in the live call signature?

PR #188, opened 2017-05-25
Entreprise support

abe33:


jlozano:
@abe33 I think this is ok to merge in since we want to get the mdfind changes in, we can test with the enterprise app next week

Issue #195, opened 2017-06-01
Switch default option for kite tour


jlozano:
@abe33  can we switch the default option to be to not show the kite tour on startup?

abe33:
Sure, it's just a matter of changing a setting's default

PR #198, opened 2017-06-01
Change kite tour display default

abe33:
Ref #195


jlozano:
@abe33 adding our slack discussion here:  can we only show the kite tour after the plugin is first installed and by default do not show the kite tour on startup?

Issue #200, opened 2017-06-02
Design pass 2


jlozano:
@abe33 great work on the first design pass! We did another pass today, let me know if anything doesnt make sense!

## Expand Panel
- [ ] 1) Sticky headers
  - [x] 1a) (TODO @abmirayo to add mockups)
    - Update: https://www.dropbox.com/s/s0oaym290d4qbjx/PluginsAudit_April2017Rev3.jpg?dl=0
  - [ ] 1b) Decrease spacing between headers so that you do not see text and whitespace flashing between headers as you scroll.

- [x] 2) Usages
  - [x] 2a) Switch title from `Usages` to `Usages from your code`
  - [x] 2b) Put usages above examples

## Live signature panel
- [x] 1) The default arguments are not being shown correctly (TODO @dhung09) (see `b` keyword arg in gif):
![atom-should-highlight-kwargs-index](https://cloud.githubusercontent.com/assets/425691/26711953/ce49262e-4718-11e7-99ae-f3693f41a76b.gif)

- [x] 2) Set max width of panel to 30 em
- [x] 3) Wrap popular patterns by token instead of adding scrollbar

- [x] 4) `**kwargs`
  - [x] 4a) Can't click on the`**kwargs` link when you are typing an argument:
![atom-patterns-unresponsive](https://cloud.githubusercontent.com/assets/425691/26711490/7007839c-4715-11e7-80cb-b82e32e74a8c.gif)
  - [x] 4b) Once `**kwargs` are expanded you cannot type anymore
  - [x] 4c) If the user is typing an argument that matches a `**kwarg` then expand the `**kwargs` section and highlight the argument. This information is included in the response, (note that the `in_kwargs` field is true and the index is relative to the `**kwargs` not the standard args when the user is typing an argument that matches a `**kwarg`):
![atom-should-highlight-kwargs-index](https://cloud.githubusercontent.com/assets/425691/26711467/4c0fbc5c-4715-11e7-8bc1-9283fe0417c5.gif)
   - [x] 4d) Popular patterns disappear when typing an argument (looks like we maybe switch to completions?): 
![atom-patterns-unresponsive](https://cloud.githubusercontent.com/assets/425691/26712098/c7ecc6e0-4719-11e7-9a0a-9884d4c5c0df.gif)

## Status panel
 - [x] 1) Always show invite others call to action in the [menu section]( https://kite.quip.com/rVcJA1bIqZny#QcWACArvrDW)

[...]

PR #201, opened 2017-06-02
Design Pass 2

abe33:
Ref #200


jlozano:
@abe33 Sitcky headers, and **kwargs changes look good! I was not able to test the kite status panel changes because nothing happens when I click on the kite status icon?

Noticed one minor thing with the signatures panel:
- [ ] 1) Is it possible to disable completions while the user is typing a function argument? It is a bit jarring:
completions-during-argument

abe33:
Hi @jlozano, so yes it's feasible, but I'm a bit concerned by hijacking too much autocomplete-plus, I mean, it's a radical change, I don't for python devs, but I often find myself relying on the completions when typing arguments of a function, especially when passing variables I just defined to a function call.

That's what led us to hide patterns in the first place, and also to reduce the amount of displayed completions so that the signature stays visible.

Now, maybe it's time to look at a different strategy, perhaps making the signature a separate panel and use a mutation observer to track the change of position of the autocomplete panel, the idea would be to behave as VSCode, a panel above the current line with the signature and one below for the suggestions. And using the observer we'll be able to handle when the suggestions gets above the current line (if there's not enough space below).

abe33:
kite-signature-v2

@jlozano I've played a bit with the idea of separating the signature from autocomple-plus, so far it seems to not be as bad as I thought initially, but there's still a few issues: 

- As I'm placing the panel over the line, we start having issues when the line is too high in the file and there's not any space above for the whole panel to fit. This is especially true when there's popular and/or that the kwarg section is expanded. Maybe a solution is to change the layout of the signature panel.
- Observing the `autocomplete-plus` panel changes is a bit harder than anticipated, mostly because the panel itself don't change, it's its parent decoration that handle the position changes, but that decoration is destroyed/recreated each time, which makes it difficult to track.

jlozano:
As discussed in real life, we will figure out the completions issue in a separate branch.

jlozano:
@abe33 looking good! some final notes:
- [ ] 1) For the invite call to action (https://kite.quip.com/rVcJA1bIqZny#QcWACArvrDW) the logic does not work as expected: we only show the "Get free Pro!" call to action 
image
when `plan.referrals_credited > 0` ? this is not quite correct we should always show the above call to action whenever `plan.referrals_credited < plan.max_referral_credits`. Let me know if this makes sense!

- [ ] 2) The Kite Status Icon is not red when kited is not running.

- [ ] 3) The Kite Status Icon is not red when the user is not logged in.

abe33:
For the last two points I'm not sure what happened, I'm seeing it as red in both case: 

kite-status_less_ ___github_kite

sample_py_ ___github_kite

abe33:
For the first point, I think it's ok now, I'm must have been confused somewhere because I was only making the switch for community plan, now this is true both for community and pro plan.

lib/elements/kite-status-panel.js
```diff
@@ -215,29 +215,32 @@ class KiteStatusPanel extends HTMLElement {
   renderLinks(account) {
     let giftLink = '';
 
-    if (!Plan.isActivePro()) {
+    if (!Plan.isEnterprise()) {
       if (Plan.referralsCredited() &&
```
jlozano: Hmm not a js expert but won't this evaluate to `false` when `Plan.ReferralsCredited() == 0`, I tried the following experiment in node:
screen shot 2017-06-14 at 3 30 47 pm

Or did i miss something?

jlozano:
@abe33 I am still seeing point 1) above:

- [ ] User is on trial pro plan with no referrals credited (see comment in pr):
```
{
  "name": "pro",
  "status": "trialing",
  "active_subscription": "pro",
  "features": {
    ...
  },
  "trial_days_remaining": 7,
  "started_kite_pro_trial": true,
  "referrals_credited": 0,
  "max_referral_credits": 10,
  "referral_days_credited": 0
}
```
- [ ] User is on pro plan with no referrals credited:
```
{
    "name": "pro",
    "status": "active",
    "active_subscription": "pro",
    "features": {
      ...
    },
    "trial_days_remaining": 0,
    "started_kite_pro_trial": true,
    "referrals_credited": 0,
    "max_referral_credits": 10,
    "referral_days_credited": 0
}
```
- [ ] User is on a community plan with no referrals credited:
```
{
  "name": "community",
  "status": "active",
  "active_subscription": "community",
  "features": {
    ...
  },
  "trial_days_remaining": 0,
  "started_kite_pro_trial": true,
  "referrals_credited": 0,
  "max_referral_credits": 10,
  "referral_days_credited": 0
}
```

jlozano:
@abe33 re the last 2 points:

It appears to have something to do with switch files or opening a new file. If I quit kite or log out of kite then the icon only changes color when I switch files? Do I need to wait a certain amount of time or do we only check when the user switches files?

screen shot 2017-06-14 at 3 37 04 pm
screen shot 2017-06-14 at 3 41 15 pm

lib/elements/kite-status-panel.js
abe33: No, you're right, it's my fault for not reading your comment properly, I missed the part when you said that it wasn't not correct to test `referrals_credited > 0`. I'll remove that condition.

abe33:
Re last two points, we didn't put any polling for the status, the check is made only when switching file, but I can add some light polling, something like every 15s or 30s.

PR #204, opened 2017-06-08
Javascript support

abe33:
For the moment this PR includes the changes made in the `design-pass-2` branch, this won't be an issue given the other PR should have been merged before this one.


jlozano:
@abe33 can we merge this in? would be good to start using the new api fields so that we can remove the deprecated fields from the response. Even though JS support is not complete.

jlozano:
@abe33 feel free to merge when ready!

PR #207, opened 2017-06-13
Swap order of signatures and completions in the overlay

abe33:


jlozano:
sorry for the delay on this, LGTM!

Issue #242, opened 2017-09-08
"more" link in hover on instance renders the instances type


jansorg:
Hovering on img (line 3) displays a hover for the instance img (correct). Clicking "more" opens the type None, though. I think that the instance itself should be rendered.

```python
class MyClass(object):
    def __init__(self):
        self.img = None
```

![instance-hover](https://user-images.githubusercontent.com/11473063/30219832-96c38b34-94bd-11e7-880b-b0b53af4f8db.png)

abe33:
Hi @jansorg, That's because I'm using the value endpoint and not the symbol one. I've started a branch to change that but I see some inconsistencies in responses (missing examples, links, etc.) for some kind of data (mostly modules).

abe33:
One good example of the missing data is for the `json` module. Using the value endpoint we get examples, SO links, and a much longer summary with code examples. The symbol endpoint, on the other hand, only returns a really short description with no examples nor SO links.

PR #243, opened 2017-09-11
Use symbol endpoint in sidebar

abe33:
Ref #242


jansorg:
@abe33 I tested this on Windows and it looks good to me! Feel free to merge...

PR #254, opened 2017-09-25
Update status

abe33:


jansorg:
@abe33 Thanks! This is looking good :)

A few things I noticed:
- The panel should show the text "The current file is ignored by Kite". It currently displays "Kite engine is not enabled for this file". See https://kite.quip.com/rVcJA1bIqZny/Kite-Status-Spec#QcWACAOmOsW
- The little dot is missing if the current file is not supported by Kite
atom-status-unsupported-file

jansorg:
@abe33 I tested the new changes, it's looking good!
Feel free to merge...

PR #259, opened 2017-10-04
Fix examples input

abe33:
It introduces a mechanism similar to what we're using for the outputs

This gives this kind of results: 

sample_py_ ___github_kite


lib/elements/kite-curated-example.js
```diff
@@ -143,13 +162,14 @@ class KiteCuratedExample extends HTMLElement {
     })
     .join('');
     const inputFiles = data.inputFiles;
+    console.log(inputFiles);
```
jansorg: Do we wan't to keep this debug logging?

lib/elements/kite-curated-example.js
```diff
@@ -105,6 +105,20 @@ const OUTPUTS = {
     <pre class="input"><code>${escapeHTML(atob(part.content.data))}</code></pre>`,
 };
 
+const INPUTS = [
+  [
+    p => p.mime_type === 'image/png',
```
jansorg: I can't find a spec with the list of mime types. I guess that `image/jpeg` could be common, too.

PR #260, opened 2017-10-04
Fix file path in urls on windows

abe33:


abe33:
@jansorg I have added tests for that, I'll take some time to add more tests for the other cases later as this is not reall the scope of this PR

jansorg:
Thanks, Cédric!
I think I tested this already before the tests were added. Feel free to merge this into master.

PR #283, opened 2017-11-22
Implement kited health segment metrics

abe33:
This PR implements the `kited_health` metric sent to Segment. 

The switch between development and production mode is done using the `inDevMode` helper of the Atom APIs. 

No events are sent when running tests.

The `userId` field can't be `0` as it seems the nodejs lib is validating it, `'0'` makes the validation pass.

### DIFF
--- a/lib/kite.js
+++ b/lib/kite.js
@@ -309,6 +309,10 @@ module.exports = {
       }
     }, atom.config.get('kite.pollingInterval'));
 
+    // We monitor kited health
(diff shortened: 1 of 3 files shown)


jansorg:
@abe33 Thanks, the properties are looking fine now on segment.com.
Is it possible to use the mac-based ID as specified at https://kite.quip.com/0asAAfQYzfNg/Editor-Sourced-Metrics#VRGACAmMwEQ ?

PR #284, opened 2017-11-27
UI changes

abe33:
As described in kiteco/kiteco#5075

### DIFF
--- a/lib/elements/html-utils.js
+++ b/lib/elements/html-utils.js
@@ -75,7 +75,7 @@ function renderValueHeader(symbol) {
 function renderHeader(name, type) {
   return `
   <div class="expand-header split-line">
-    <span class="name">${name}</span>
+    <span class="name">${name.replace(/\./g, '.&#8203;')}</span>
     <span class="type">${type}</span>
   </div>`;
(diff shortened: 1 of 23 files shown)


jansorg:
@abe33 
- [ ] I've noticed that the login form displays both "SIgn into Kite" and "Create account" as label of the first button of the login dialog. screenshot_20171204_180706
- [ ] I logged out of Kite and opened the login box. After a few seconds the "Login" notification appeared in the background. I think the notification shouldn't be shown while the login box is in the foreground (see the screenshot above)
- [ ] The first letter should be uppercase in  "user with email ... already exists.
screenshot_20171204_181031
- [ ] Search: List of parameters cut off. There's cut-off content between the last visible parameter "alpha" and the heading "Docs". The content of "Docs" is cut off later on, as well.
screenshot_20171204_181838

abe33:
> he first letter should be uppercase in "user with email ... already exists.

The message actually comes from kited, not sure that this is something we should handle on a extension by extension basis

abe33:
The list cutoff was due to collapsible section being broken.

PR #307, opened 2018-02-12
Fixes batch

abe33:


abe33:
Re v0 vs v1, the reason was that when we started writing that plugin v1 wasn't implemented yet, and atom's electron version is always a bit behind (currently, atom's master is running Chromium 58). If we still want to support older version of Atom we can't move to v1 yet as the Chromium version they ran might be lower that 54.

PR #312, opened 2018-02-15
Autocorrect improvements

abe33:
This PR covers the support for hash mismatch feedback as well as the thumb up/down feedbacks.

### DIFF
--- a/lib/data-loader.js
+++ b/lib/data-loader.js
@@ -3,9 +3,10 @@
 const {StateController, Logger} = require('kite-installer');
 const {
   accountPath,
-  autocorrectPath,
-  autocorrectMetricsPath,
   authorizedPath,
+  autocorrectFeedbackPath,
+  autocorrectMetricsPath,
+  autocorrectPath,
   completionsPath,
   examplePath,
   hoverPath,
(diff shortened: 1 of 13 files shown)


jansorg:
@abe33 Thanks, this is looking good! :)

- The diff sections are a bit hard to read, some more contrast would be great.
- I clicked on thumbs-up but wasn't sure if it was successful because nothing happened. Would it be possible to add some ui feedback on mouse-over and possible on click (e.g. changing color or something)?
- Would it be possible to render the time in local time?
- @dhung09, @jlozano Should be it possible to submit up/down feedback more than once for a given diff?

This is how it's rendering on Win10 with the default theme (I assume):
screenshot_20180219_162735

jansorg:
@abe33 Thanks, this is much better :)
The line numbers in the left part of the diff are still hard to read, is it possible to fix these as well?

jlozano:
@jansorg good question! We should only allow users to submit feedback once per set of diffs. @abe33 can we gray out or disable the feedback buttons once they submit it once?

abe33:
@juan ;)
autocorrect_improvements_by_abe33_ _pull_request__312_ _kiteco_atom-plugin

jlozano:
@abe33 I updated the spec with a couple of changes to make it consistent with the other editor metrics:
- There is now a `meta` field for all requests, that includes the following information `source`, `event`, `os_name`, `plugin_version`, see https://kite.quip.com/FgQBA4MESbWi/Autocorrect-API#ecPACA2PkU7.
- I removed the `editor` field since it is now included as part of the `meta` object.
- I added a `response_time` field to the hash mismatch request, is it easy to measure this from the plugin end?

abe33:
@jlozano last commit adresses the spec changes. As for the `response_time` I diff from the moment I call the first endpoint to the moment I detect the hash mismatch, I think it should be fine enough for a first version as I don't think md5 computation is so costly that it will make the measure irrelevant.

PR #315, opened 2018-02-19
Change status settings url

abe33:
Replace all calls to `http://localhost:46624/settings` and sub routes with calls to `kite://settings`.


lib/elements/kite-localtoken-anchor.js
```diff
@@ -8,6 +8,7 @@ class KiteLocaltokenAnchor extends HTMLAnchorElement {
     });
   }
   createdCallback() {
+    console.log('here', this.getAttribute('href'));
```
dbratz1177: Take out this log?

PR #317, opened 2018-02-21
Replace install flow with v1 version

abe33:


adamsmith:
Hi all, should we merge this?

Issue #319, opened 2018-02-22
Atom Kite Startup Time


arv1362:
The startup time of kite is increasing as time goes on which slows my coding time till i have to disable the package

![capture](https://user-images.githubusercontent.com/33993859/36541205-21a283f8-1818-11e8-8e31-34cae0c2e2bb.PNG)

abe33:
Hello @arv1362,

Could you tell us a few things regarding your config?

- What is your OS?
- What is your Atom version?
- Do you see the same long startup time when the kite process is not running than when it is running?
- Do you see the same long startup time when starting Atom with no project?
- Do you see the same long startup time when starting Atom in a new directory?

arv1362:
-Windows 7 Ultimate
- Version 1.24.0
-The startup time is slightly shorter (3440ms) when the kite is not running
-No, the startup time will be very short
-Yes

arv1362:
Windows 7 Ultimate

Version 1.24.0
-The startup time is slightly shorter (3440ms) when the kite is not running
-No, the startup time will be very short
-Yes

arv1362:
Windows 7 Ultimate

Version 1.24.0
-The startup time is slightly shorter (3440ms) when the kite is not running
-No, the startup time will be very short
-Yes

abe33:
Thanks @arv1362, that helps a lot! I might have figured a way to speed up the package startup time and I'll see with the other developers in charge of the kite process how to handle this.

abe33:
Maybe a last question. 

Do you see a difference in startup time when there are files open in a project VS when no files are open?

arv1362:
It doesn't show any significant difference even when no files are open

abe33:
Hey @arv1362, we've released a version that should have improved things, can you confirm it or do you still see this high activation time?

jansorg:
@arv1362 If the newest version is still too slow at startup, then let us know, please.

PR #320, opened 2018-02-22
Do not return the connection promise in activate

abe33:
It should help reducing the activation time if the process to connect to kited takes more time than usual.

Ref #319

### DIFF
--- a/lib/kite.js
+++ b/lib/kite.js
@@ -436,7 +436,7 @@ module.exports = {
     }
 
     // We try to connect at startup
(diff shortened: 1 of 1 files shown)


abe33:
Yes, basically the package will only be considered as activated once the returned promise is resolved, it also means that if the promise fails the package activation will fail as well. This is also useful in tests as you can ensure that the initial processing done during activation are accounted for when waiting for the plugin activation in your before hook.

PR #327, opened 2018-03-06
Use position instead of range in hover calls

abe33:


jansorg:
@abe33 I think there's still a hover request which uses the range in `lib/urls.js`

abe33:
@jansorg It's not a request it's just that I haven't removed the helper that build the url with a range, it is actually no longer used

abe33:
@dbratz1177 Yeah totally, in that case we need the range to create a text decoration for the word to show the underline on hover. We're not interacting with the hover endpoint at this point, we're still using the tokens retrieved from the `tokens` endpoint to checks whether we can support hyperclick here or not.

abe33:
@jansorg That's probably in the case of the expand panel where we're requesting the hover endpoint beforehand to the symbol endpoint, not sure this is concerned by the change, do you have any info on it?

abe33:
@jansorg That's probably in the case of the expand panel where we're requesting the hover endpoint beforehand to the symbol endpoint, not sure this is concerned by the change, do you have any info on it?

PR #333, opened 2018-03-12
Remove events for unsupported files

### LINKED ISSUE #331: Remove events for non-python/unsupported files
Events for unsupported files should be removed. 
Also, there should be no requests send to kited for these files (completions, ...)
The implementation to identify supported files should not change.

cc @tarakju

abe33:
Closes #331

### DIFF
--- a/lib/completions.js
+++ b/lib/completions.js
@@ -20,22 +20,25 @@ const KiteProvider = {
 
   // called to handle attribute completions
   getSuggestions(params) {
-    return delayPromise(() => {
-      let promise = Plan.queryPlan();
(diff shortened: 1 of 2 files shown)


dbratz1177:
FYI (perhaps stating the obvious) - in the conflict resolution, we want to keep the `from` parameter being passed to `check`

dbratz1177:
My testing results:
- when logged-in: 
    - no endpoint hits for non-python and JS files
    - endpoint hits for JS files (`event`, `completions`, and `signatures`)
- when not logged-in:
    - no endpoint hits for non-python and JS files
    - `event` endpoint hit for both JS and python files

PR #336, opened 2018-03-14
Switch useDockForSidebar to true by default

abe33:


dbratz1177:
Something I noticed in this PR - not sure if its handled elsewhere - but that if I do:
1. hover over some token, and get the `web / more` popup
2. click 'more'
3. hit the `>` on the dock sidebar
4. repeat 1 and 2
Then then dock remains closed (I think it should reopen - ala what the previous sidebar version did)

PR #337, opened 2018-03-14
Remove legacy sidebar implementation in favor of the dock API

abe33:
This PR basically remove the setting and use the dock API version wherever we were switching behavior based on that setting. Also, the sidebars no longer have a close button nor a resizer as we're letting the dock handles that for us.

### DIFF
--- a/lib/elements/kite-autocorrect-sidebar.js
+++ b/lib/elements/kite-autocorrect-sidebar.js
@@ -39,11 +39,7 @@ class KiteAutocorrectSidebar extends HTMLElement {
 
   createdCallback() {
     this.innerHTML = `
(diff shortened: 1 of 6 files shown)


jansorg:
@abe33 I noticed that the plugin's sidebar is no longer adjusting when resized.
screenshot_20180314_175643

jansorg:
Clicking the X in the upper right corner is not closing the panel.

dbratz1177:
I'm also seeing the behavior @jansorg sees

abe33:
Yeah sorry forgot to remove the close button and fixed width part, fixed now

mwichary:
👍

PR #338, opened 2018-03-14
Remove use of tokens endpoint

abe33:
Ref kiteco/kiteco#5451

- [x] remove tokens endpoint calls
- [x] use hover call for documentation hint in status bar

I think hyperclick and context menus are too complex to solve right now, it'd be preferable to remove them and ship these changes and then devise a new strategy to put them back.

### DIFF
--- a/lib/data-loader.js
+++ b/lib/data-loader.js
@@ -12,7 +12,6 @@ const {
   hoverPath,
   hoverPositionPath,
   internalExpandURL,
-  internalGotoIdURL,
   languagesPath,
   membersPath,
   onSaveValidationPath,
@@ -22,7 +21,6 @@ const {
   signaturePath,
   statusPath,
   symbolReportPath,
-  tokensPath,
   usagePath,
   usagesPath,
   valueReportPath,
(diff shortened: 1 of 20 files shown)


dhung09:
@abe33 I think we can go ahead and remove the context menus and hyperclick functionality for now, so feel free to do that in this PR if not done already (cc @tarakju)

abe33:
Yes, I removed everything that worked on tokens so hyperclick and context menu are no longer available with this PR.

tarakju:
@abe33 When using this branch, I noticed the following calls:
```
[kited] 2018/03/21 20:12:36.322600 handlers.go:64: GET /api/buffer/atom/:Users:tarak:go:src:github.com:kiteco:kiteco:kite-python:kite:codeexamples:inference.py/d38d6059db46de6ad660c1e3fd13e2e1/hover?selection_begin_runes=437&selection_end_runes=452 200 5519 37.800205ms
```

Seems like its using the old query parameters for `hover`.

dbratz1177:
wrt to the old hover - just having searched through the code - there's a few refs to `kiteEditor.expandRange()` sprinkled through the code, which routes to `showDataAtRange` in `navigable-panel.js`, which, through `getReportDataAtRange`, calls the old hover endpoint

I suspect that's what caused the behavior @tarakju was talking about

tarakju:
@abe33 @dbratz1177 I think we're going to handle the hover issue separately.

PR #341, opened 2018-03-15
eliminates .js files from consideration for suggestions

dbratz1177:
quick fix to issue with js files triggering calls to `clientapi/signatures` and `clientapi/completions`
Should probably, long-term, use similar `isSupported` logic (ala #333) to wrap the internals of `getSuggestions` call.


abe33:
I've made the change in my PR to use the `isGrammarSupported` method in `getCompletions`, not sure whether we should make the change in the declared scope as well, but, anyway, this is done now.

PR #346, opened 2018-03-15
Makes autocorrect pane use user’s code fonts.

mwichary:
@abe33 /cc @dbratz1177 

This is a pretty naïve way of doing this. I can imagine in the future this needs to be extended to cover all the panels (and also react in real time when user makes changes to these fonts).

However, I’m also interested in getting this in for autocorrect first, to see if we get any feedback on this.

Let me know if this needs to be structured better or placed elsewhere even for MVP.

Before, generic font:
screen shot 2018-03-15 at 10 18 26

After, exactly the font I use for editing:
screen shot 2018-03-15 at 11 09 55
image

### DIFF
--- a/lib/elements/kite-autocorrect-sidebar.js
+++ b/lib/elements/kite-autocorrect-sidebar.js
@@ -37,23 +37,8 @@ class KiteAutocorrectSidebar extends HTMLElement {
     return atom.config.get('kite.sidebarWidth');
   }
 
-  // Grabs code typography styles from Atom configuration and turns
-  // them into a CSS block
-  getTypographyStylesFromConfig() {
-    return `
-      <style>
-        code {
-          font-family: ${atom.config.get('editor.fontFamily')};
-          font-size: ${atom.config.get('editor.fontSize')}px;
-          line-height: ${atom.config.get('editor.lineHeight')};
-          tab-size: ${atom.config.get('editor.tabLength')};
-        }
-      </style>`
-  }
-
   createdCallback() {
     this.innerHTML = `
-      ${this.getTypographyStylesFromConfig()}
       <div class="kite-sidebar-resizer"></div>
       <div class="kite-column">
         <header>
--- a/lib/elements/kite-autocorrect-sidebar.js
+++ b/lib/elements/kite-autocorrect-sidebar.js
@@ -37,8 +37,23 @@ class KiteAutocorrectSidebar extends HTMLElement {
     return atom.config.get('kite.sidebarWidth');
   }
 
+  // Grabs code typography styles from Atom configuration and turns
+  // them into a CSS block
+  getTypographyStylesFromConfig() {
+    return `
+      <style>
+        code {
+          font-family: ${atom.config.get('editor.fontFamily')};
+          font-size: ${atom.config.get('editor.fontSize')}px;
+          line-height: ${atom.config.get('editor.lineHeight')};
+          tab-size: ${atom.config.get('editor.tabLength')};
+        }
+      </style>`
+  }
+
   createdCallback() {
     this.innerHTML = `
+      ${this.getTypographyStylesFromConfig()}
       <div class="kite-sidebar-resizer"></div>
       <div class="kite-column">
         <header>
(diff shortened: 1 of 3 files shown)


lib/elements/kite-autocorrect-sidebar.js
```diff
@@ -37,8 +37,23 @@ class KiteAutocorrectSidebar extends HTMLElement {
     return atom.config.get('kite.sidebarWidth');
   }
 
+  // Grabs code typography styles from Atom configuration and turns
+  // them into a CSS block
+  getTypographyStylesFromConfig() {
```
abe33: We don't need to do that, actually Atom is already doing the same except for the tab length (but that should be irrelevant for us here as our diffs should always be one liner, and initial tabbing is probably not desired)

Basically Atom is creating and updating 3 css variables `--editor-font-family`, `--editor-font-size` and `--editor-line-height` that we can use in our stylesheet like this: 

```css
font-family: var(--editor-font-family);
font-size: var(--editor-font-size);
line-height: var(--editor-line-height);
```

abe33: That also means we can use that for code block in the docs panel as well.

styles/kite-autocorrect-sidebar.less
```diff
@@ -67,11 +67,11 @@ kite-autocorrect-sidebar {
     .line-number {
       width: 80px;
       opacity: 0.5;
-      padding: @component-padding;
+      padding: 4px @component-padding;
```
abe33: I'm not sure that hard-coded values makes sense here, if we're using the `@component-padding` variable it's because themes can have different defaults for this. If we need to have subtle adjustments I've already created derived variables for that (such as `@half-padding`, `@double-padding` and `@triple-padding`), `@half-padding` is `5px` on the default theme so I think it's close enough that we don't need to hardcode anything.

lib/elements/kite-autocorrect-sidebar.js
mwichary: Thanks! This, however, doesn’t work for me if I copy and paste your code, and I once again I’m having zero luck in finding documentation for it. :·/ Maybe you could point me in the right direction?

Agree about tabs.

styles/kite-autocorrect-sidebar.less
mwichary: Hmm. Thanks for the context. Thinking about it, this one should be a function of font size, not theme. I can change it to be expressed in `em`s. Should I create a variable for that, too?

mwichary:
@abe33 Added all of my comments, please take a look again?

lib/elements/kite-autocorrect-sidebar.js
abe33: Hmm right... This hasn't landed in a release yet... (I'm using a master built version to catch changes early). 
That's a problem, we probably don't want to implement something that will be obsolete in a few versions because a native mecanism that does the same thing will be available. My biggest concern with this approach is the way you're doing it you change every `code` element in Atom, not just the one we're using. Ideally we could have used `scoped` css but this has been dropped from the specification process and no support exist for that (https://www.caniuse.com/#search=scoped). That left us with the solution of creating these variables instead of styling the `code` element and maybe add an existence check to make to not create them when they'll become available natively. 
Also, if we're going with that route, this change should probably not be made in just one component whose lifecycle in not known before hand. If we want to have these styles consistently we must put that at the plugin activation time. I guess we can just include that helper function from the upcoming Atom version (https://github.com/atom/atom/blob/e3a464eadf0d8fc8d4c2cac4506f69a1f5c50243/src/workspace-element.js) and make the existence test based on that method name

abe33: Hope this make sense.

mwichary: I can just scope it better with `kite-autocorrect-sidebar code {`? (Also, I haven’t seen it affecting anything else FWIW.)

Also, I definitely agree there is a better place for it – but I thought this would be a nice self-contained test where we could catch people’s complaints or things looking weird before we expand it to the main sidebar and/or move to native variables when they become available? A little typography MVP.

styles/kite-autocorrect-sidebar.less
abe33: If you intend to consistently use these variables accross all UIs sure it makes sense to store that in a variable.

mwichary:
@abe33 Will you take the ownership of this, or should I do anything else here?

PR #350, opened 2018-03-15
Better feedback.

mwichary:
@abe33

· replace tooltips when you hover over thumbs up/down to be nicer and quicker
· add post-click feedback, too
· move thumbs up/down to the help
· style thumbs up/down to be less prominent

screen shot 2018-03-15 at 14 23 47
screen shot 2018-03-15 at 14 23 40

### DIFF
--- a/lib/elements/kite-autocorrect-sidebar.js
+++ b/lib/elements/kite-autocorrect-sidebar.js
@@ -62,14 +62,14 @@ class KiteAutocorrectSidebar extends HTMLElement {
   }
 
   // Show hint or confirmation when the user hovers or clicks thumbs up/down
-  showFeedbackFeedback(e, confirmation, message) {
-    let el = parent(e.target, '.feedback-actions').querySelector('.feedback-feedback')
+  showFeedbackFeedback(target, message, {confirmation}={}) {
+    let el = parent(target, '.feedback-actions').querySelector('.feedback-feedback')
     el.classList.add(confirmation ? 'confirmation' : 'hint')
     el.innerHTML = message;
   }
 
-  hideFeedbackFeedback(e) {
-    let el = parent(e.target, '.feedback-actions').querySelector('.feedback-feedback')
+  hideFeedbackFeedback(target) {
+    let el = parent(target, '.feedback-actions').querySelector('.feedback-feedback')
     el.classList.remove('hint')
   }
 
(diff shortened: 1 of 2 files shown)


lib/elements/kite-autocorrect-sidebar.js
```diff
@@ -61,6 +61,18 @@ class KiteAutocorrectSidebar extends HTMLElement {
     this.element = this;
   }
 
+  // Show hint or confirmation when the user hovers or clicks thumbs up/down
+  showFeedbackFeedback(e, confirmation, message) {
+    let el = parent(e.target, '.feedback-actions').querySelector('.feedback-feedback')
```
abe33: I think we should just pass the target instead of the event in that case, here we're making that function only callable from within an event listener.

abe33: Also, but that's not too much of an issue here, lately people are considering using boolean arguments defined like that as a bad argument, mostly because when you see a call such as `doSomething(2, true, false, false)` that doesn't make any sense. The solution to this is to use deconstruction on an option object: 
```js
function doSomething(arg1, {myBool1, myBool2, myBool3}={}) {
   ...
}

doSomething(2, {myBool1: true, myBool2: false, myBool3: false})
```

That way it's easier to make explicit what each bool is refering to and it also makes it easier to omit some of them

lib/elements/kite-autocorrect-sidebar.js
```diff
@@ -90,12 +102,32 @@ class KiteAutocorrectSidebar extends HTMLElement {
...
+
+    this.subscriptions.add(addDelegatedEventListener(this, 'mouseover', '.thumb-up', (e) => {
+      this.showFeedbackFeedback(e, false, 'Send feedback to Kite if you like this change')
+    }));
+
+    this.subscriptions.add(addDelegatedEventListener(this, 'mouseover', '.thumb-down', (e) => {
+      this.showFeedbackFeedback(e, false, 'Send feedback to Kite if you don’t like this change')
+    }));
+
+    this.subscriptions.add(addDelegatedEventListener(this, 'mouseout', '.thumb-up', (e) => {
```
abe33: Here our `addDelegatedEventListener` allows to listen to several events or several targets, so it's possible to write this instead: 

```js
this.subscriptions.add(addDelegatedEventListener(this, 'mouseout', '.thumb-up, .thumb-down', (e) => {
  this.hideFeedbackFeedback(e)
}));
```

styles/kite-autocorrect-sidebar.less
```diff
@@ -48,17 +48,37 @@ kite-autocorrect-sidebar {
...
 
   .feedback-actions {
-    text-align: right;
+    display: flex;
+    text-align: left;
     margin-top: @component-padding;
 
     a {
-      margin-right: @component-padding;
+      padding-left: @component-padding / 2;
```
abe33: We have a `@half-padding` variable if you don't want to make that computation every time.

mwichary: Nice.

lib/elements/kite-autocorrect-sidebar.js
mwichary: Oh, man, I love that convention. I started using something similar to this years ago on a project.

mwichary:
@abe33 Made all the requested changes, please re-review?

PR #351, opened 2018-03-15
Remove test-autocorrect-ui

mwichary:
@jlozano

I assume we want to do this before launch?


jlozano:
@mwichary good question, I do not know, @abe33 would have more context on this?

abe33:
Yeah sure it was there to work on the styles without having to constantly save/checkout a file. I can just make it available when starting Atom in `--dev` mode.

mwichary:
@abe33 Sounds good, should I close this?

PR #354, opened 2018-03-19
New autocorrect api support

abe33:
This is what you get when running the autocorrect UI command. It's all the stuff grouped together for styling purpose. 
kite-autocorrect-sidebar_js_ ___github_kite


lib/kite.js
```diff
@@ -142,8 +142,7 @@ module.exports = {
       // this.app.saveUserID();
     }));
 
-    const workspaceElement = atom.views.getView(atom.workspace);
-    if (!workspaceElement.updateGlobalTextEditorStyleSheet) {
+    if (parseFloat(atom.getVersion()) <= 1.25) {
```
mwichary: Oh, no. 😬

dbratz1177:
TIL `atom -d` does not implicitly run an `npm install`

lib/elements/kite-autocorrect-sidebar.js
```diff
@@ -93,42 +126,65 @@ class KiteAutocorrectSidebar extends HTMLElement {
...
       DataLoader.postAutocorrectFeedbackData(this.getCurrentEditorLastCorrections(), +1);
       e.target.classList.add('clicked');
       parent(e.target, '.diff').classList.add('feedback-sent');
 
-      this.showFeedbackFeedback(e.target, 'Thanks for your feedback!', {confirmation: true})
+      this.showFeedbackFeedback(e.target, 'Thanks for your feedback!', {confirmation: true});
     }));
 
     this.subscriptions.add(addDelegatedEventListener(this, 'click', '.thumb-down', (e) => {
       DataLoader.postAutocorrectFeedbackData(this.getCurrentEditorLastCorrections(), -1);
```
dbratz1177: Do we want to allow the user to spam feedback hits?

dbratz1177:
Maybe a `Go to code` link for `Earlier Corrections` too?

dbratz1177:
This is a comment that is probably higher level than this PR, and maybe plans are already in the works to address it, but when I click on the link to `help.kite.com`, as a user, I would expect to be taken to a link with a clear affordance/pointer to something specific about Autocorrect.
The current `help.kite.com` leaves one hanging in that regard

lib/kite.js
abe33: Yeah, but we don't really have much option here, there's no `capabilities` API that would allow us to know when specific features are available, and we can't always rely on methods existence as they can exist with a different purpose (that was the case here).

abe33:
@dbratz1177 We've put the link here as a placeholder for when we'll have the actual code clean-up url.

PR #356, opened 2018-03-21
Last batch of UI fixes.

mwichary:
@abe33

· baseline alignment for options
· thumbs up/down from Octoicons, better visibility on light bk
· swapped question mark for info
· better highlight for thumbs up when clicking around
· diff style and fixed +/– positioning

screen shot 2018-03-21 at 09 53 51

### DIFF
--- a/lib/elements/kite-autocorrect-sidebar.js
+++ b/lib/elements/kite-autocorrect-sidebar.js
@@ -64,7 +64,7 @@ class KiteAutocorrectSidebar extends HTMLElement {
                 <div class="setting-title">Any time code is fixed:</div>
                 <select type="checkbox" class="form-control">
                   <option>Reopen this sidebar</option>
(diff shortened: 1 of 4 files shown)


lib/elements/kite-autocorrect-sidebar.js
```diff
@@ -63,7 +63,7 @@ class KiteAutocorrectSidebar extends HTMLElement {
                 <div class="setting-title">Any time Kite fixes code:</div>
                 <select type="checkbox" class="form-control">
                   <option>Reopen this sidebar</option>
-                  <option>Cleanup quietly</option>
```
abe33: If you change the option wording you also have to do it in every other places, the way enums works in settings is quite restrictive so we can't have a label != than the value. By other places I mean: 
- `package.json` where the authorized options are defined
- `autocorrect-spec.js` where this setting is used in the tests setup

PR #358, opened 2018-03-22
no event, completion, or sig hit for non-whitelisted files

dbratz1177:
adds checks to eliminate `clientapi/editor/[event | signature | completion]` calls for non-whitelisted python files
cc: @tarakju 
should eliminate a number of `403`s to those endpoints

NB: I added a quick and somewhat dirty hack to get the tests to pass for this. Would love to do it in a cleaner way, but this seemed by far the simplest way to do so

### DIFF
--- a/lib/editor-events.js
+++ b/lib/editor-events.js
@@ -10,8 +10,6 @@ class EditorEvents {
   constructor(editor) {
     if (!Kite) { Kite = require('./kite'); }
     this.editor = editor;
-    //this is a quick and hacky way to get the tests to pass in editor-events-spec.js
-    //would love to find a cleaner way to do this
     this.subscriptions = new CompositeDisposable();
     this.pendingEvents = [];
 
--- a/lib/editor-events.js
+++ b/lib/editor-events.js
@@ -7,24 +7,19 @@ const {last, promisifyRequest, DisposableEvent} = require('./utils');
 let Kite;
 
 class EditorEvents {
-  constructor(editor, isTest) {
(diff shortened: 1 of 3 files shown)


lib/editor-events.js
```diff
@@ -4,10 +4,14 @@ const {CompositeDisposable} = require('atom');
...
 const {MAX_FILE_SIZE, MAX_PAYLOAD_SIZE} = require('./constants');
 const {last, promisifyRequest, DisposableEvent} = require('./utils');
+let Kite;
 
 class EditorEvents {
   constructor(editor) {
+    if (!Kite) { Kite = require('./kite'); }
     this.editor = editor;
+    //this is a quick and hacky way to get the tests to pass in editor-events-spec.js
+    //would love to find a cleaner way to do this
```
abe33: If it's just a matter of having `isEditorWhitelisted` returning true in the test, you can set `Kite.whitelistedEditorIDs[editor.id]`in the before hook. I don't think the events endpoint ever returns 403 so it looks like some unecessary change in the whitelist state (every time the `handle403Response` will get a 200 the file will be considered as whitelisted, and with a 403 you start the whole whitelist/blacklist notifications stuff)

dbratz1177: I actually got rid of the code that the comment (and have subsequently gotten rid of the comment) was referencing in the latest commit - since, to follow the spirit of the whitelisting spec, it was decided to not gate `/events` on whitelisting.
Noted on the tip, though - that's good to know

PR #361, opened 2018-03-26
Swap autocorrect new model icon to blue info icon from a red exclamation point icon.

mwichary:
@abe33

image


jlozano:
@mwichary do you mind updating this in the mockups when you get a chance?

PR #364, opened 2018-03-27
More autocorrect fixes

abe33:
Addresses latest feedback in UI checklist.


mwichary:
@abe33 I love your use of emoji in commit names.

PR #371, opened 2018-03-29
Fix last use of range in hover calls

abe33:


abe33:
@dbratz1177, hmm, it does work for me in that branch VS master which is broken. Is it possible you were running master without noticing? (due to kite overriding our symlink in `~/.atom/package`)

dbratz1177:
@abe33 I just sanity checked - still nothing on master, and still nothing on this branch. I made sure I wasn't accidentally using master explicitly while testing this time.
My steps:
0. repo to this branch (I have it stored at `~/github/kite`)
1. `atom -d ~/Desktop/whitelisted-test.py`
2. file looks like: 
```
import json
json.dumps
```
3. with cursor after the 's' in 'dumps', bring up command palette with `Cmd+Shift+p`
4. Execute `Kite: Open in web`
Nothing observable (so far as I can tell) happens. It's weird that you and I are given different behavior

abe33:
kite-open-in-web

@dbratz1177 Could it be the issue comes from somewhere else? Like your kited version? For me it is working in both branches (master and this one). Also, you should be able to see a request in the devtools, maybe that request fails for whatever reason.

dbratz1177:
@abe33 So this is what's happening to me:
- When I execute `Kite: Open in Web`, I consistently hit the case with the `console.log('3')` in the code below:
screen shot 2018-04-09 at 10 12 18 am
- In the called `openInWeb` method, `step` is consistently `undefined`:
screen shot 2018-04-09 at 10 12 45 am

Hopefully the above helps to pinpoint what the discrepancy is

abe33:
@dbratz1177 oh right, it was the expected behaviour in the early specs, basically if the sidebar is open the `Open in web` command triggers the corresponding sidebar action. I don't think this should be addressed in that PR as it's unrelated to the use of range in hover calls.

PR #372, opened 2018-04-03
Update strings to match the new glossary.

mwichary:
@abe33

***PLEASE DO NOT MERGE***

Question: Is renaming commands okay? Will it have consequences if people already mapped them to other shortcuts? If so, is there a way to safely rename them?

New functionality: Re-saving some configs under new names and removing old ones. 

Otherwise just string changes.

This is part of a bigger effort documented here: https://kite.quip.com/eGu8AecvGYze/New-UI-strings-deployment


mwichary:
(Also cc @jansorg as an fyi.)

lib/kite.js
```diff
@@ -33,9 +33,18 @@ module.exports = {
       }
     }
 
-    // little thing to take out legacy config name value so it won't show up on settings
+    // Deal with legacy setting names
+    // March 2018
     atom.config.unset('kite.actionWhenKiteFixesCode');
 
+    // April 2018
+    atom.config.set('kite.searchIconPosition', atom.config.get('kite.activeSearchPosition'));
```
abe33: We probably don't want to do it that way. Currently, you're overriding whatever value the user has set, everytime, with either the old value or `undefined` after the first run of these lines. That's definitely not what we want here. 

The proper way to do that is as follow: 

```js
if (atom.config.get('kite.activeSearchPosition') != undefined) {
  atom.config.set('kite.searchIconPosition', atom.config.get('kite.activeSearchPosition'));
  atom.config.unset('kite.activeSearchPosition');
}
```

lib/kite.js
```diff
@@ -33,9 +33,18 @@ module.exports = {
...
+    atom.config.unset('kite.activeSearchPosition');
+
+    // April 2018
+    atom.config.set('kite.showWelcomeNotificationOnStartup', atom.config.get('kite.showDocsNotificationOnStartup'));
```
abe33: Same as above

lib/elements/kite-active-search.js
```diff
@@ -17,7 +17,7 @@ const GETTING_STARTED = [
 
 class KiteActiveSearch extends HTMLElement {
   static initClass() {
-    atom.commands.add('kite-active-search', {
+    atom.commands.add('kite-search', {
```
abe33: You're breaking things here, that string is the element name where these commands apply. Without changing the element name in its definition (as well as the associated styles and file name for consistency) this won't work.

mwichary: Oh, yikes. Good to know! I will redo.

lib/kite.js
mwichary: Ugh. Yes. Sorry.

mwichary:
Okay, I made the requested changes and tested search to make sure it still works.

/cc @abe33 if you want to look at it (last commit) again?

lib/elements/kite-active-search.js
mwichary: Renamed everything from active-search to search.

PR #374, opened 2018-04-05
Fix listening invalid config for editor styles

abe33:
The right names are `editor.` prefixed and not `atom.`.


mwichary:
Nice!

mwichary:
Nice!

PR #375, opened 2018-04-05
Synthesized syntax highlighting colours for low-contrast scenarios

mwichary:
@abe33 + @dbratz1177 

This is code that’s ready to go, but I’m also interested in a discussion around it.

This PR adds synthesizing a syntax highlighting palette for situations where dark UI theme is paired
with the light syntax theme, or vice versa. It does so with just CSS/LESS.

Today, this often looks bad; this PR makes it better. Compare Popular Patterns with native code:

> Left » UI theme: One Light · Syntax theme: Base16 Tomorrow Light (today, unchanged)
> Midlde » UI theme: One Dark · Syntax theme: Base16 Tomorrow Light (today, broken)
> Right » UI theme: One Dark · Syntax theme: Base16 Tomorrow Light (fixed by this PR)

image

> Left » UI theme: One Dark · Syntax theme: One Dark (today, unchanged)
> Middle » UI theme: One Light · Syntax theme: One Dark (today, broken)
> Right » UI theme: One Light · Syntax theme: One Dark (fixed by this PR)

image

I would like to consider more syntax highlighting in our panels – they look good and work better
functionally. Here’s an example of my work in progress in redesigning the panel so that there
is more syntax highlighting there.

image

Two issues:
1. I know this is tricky to get right; syntax highlighting in our panel won’t always
be perfect.
2. Likewise, this PR won’t always be able to synthesize the right colors, since it’s reading off of
variables, not actual CSS (and some colour themes don’t use variables directly).

Here are a few examples where the PR gets the colours wrong, but it still improves on the current experience contrast-wise:
image
image

Questions: How precious are programmers about the details of syntax highlighting? Is getting it right in some cases (but not all) better than nothing? Will people appreciate the benefits of syntax highlighting, even if in some cases we won’t get their colours right? 

(One way to answer this could be to deploy this PR and listen to people’s feedback!)

There are other possible solutions, listed below for reference and discussions:

```
                                                               ↓ THIS PR IS THIS SOLUTION 
----------------------------------------------------------------------------------------------------------------------
| NO SYNTAX HIGHLIGHTING      | OUR SYNTAX HIGHLIGHTING    | SYNTHESIZED                | SMART SYNTHESIZED SYNTAX   |
|                             | PALETTE                    | HIGHLIGHTING PALETTE       | HIGHLIGHTING PALETTE       |
|                             |                            |                            |                            |
| We have one dark and one    | We have one dark and one   | We use the default palette | As before, but we actually |
| light text color for        | light syntax highlighting  | and synthesized colors     | read them off of text.     |
| code. That’s it. All code   | palette. They are neutral, | from CSS variables.        | We include filters etc.    |
| looks the same.             | perhaps grayish or based   | Sometimes, the variables   |                            |
|                             | off of default code colour.| won’t be correct.          |                            |
|                             |                            |                            |                            |
| Trivial CSS                 | Simple CSS                 | More complex CSS           | JavaScript, very complex   |
----------------------------------------------------------------------------------------------------------------------
```

What do you think?


mwichary:
Thanks for the comments!

I’m also interested in the Our Syntax Highlighting Palette version. This is something we do already with the Copilot. Since we don’t really do much syntax highlighting today in the editor-naive sidebar, this might be a good way to start to see how people react to it. (I can also see this as a way to brand Kite and separate it from Atom to be a possible benefit.) We’d just need a light and dark theme.

@abe33 would you mind elaborating on the “every language/changes in the way the standard python highlighting work”? I’m not actually doing anything specific to Python here. Would love to hear your thoughts about how this could break.

abe33:
> would you mind elaborating on the “every language/changes in the way the standard python highlighting work”? I’m not actually doing anything specific to Python here. Would love to hear your thoughts about how this could break.

I'm not saying the standard python highlighting is changed by other languages, just users can use another one (ex. https://github.com/gilsondev/atom-language-django) in most case this is just a extension of the standard one, but it can happen they adds new rules with new classes (like, idk, adding support for code in string interpolation, or whatever). But we have no guarantee this is the case for all languages. The whole point of using Atom is because it's hackable to the core, so you have to expect people to hack it. 

Also, whenever you use `.syntax--python` you're coupling these styles to the python language and the way the highlighter works at the moment. We've strived to make the plugins unaware of the languages supported by Kite so this is going against the effort we've made so far. Additionally, in almost a 1 year time span, Github changed the way the highlighter and themes works twice (making shadow DOM the default and then removing it and introducing these `syntax--` prefixes on every highlighting classes). They're also working on a new version of the editor part that would simply drop the DOM completely (https://github.com/atom/xray). This mean that we're adding a 

I'm not saying we shouldn't use the highlighter for our code sample, I'm saying that we have no data backing up that it's worth to actually generate a whole new set of colors for a single language (meaning this work has to be done for each new language we want to support, or whenever Github decides to change things up in the highlighter) just to make things *prettier* for an unknown percentage of the user base. Like I said, the first thing would be to at least try to see how many of our users are actually using a light/dark themes combination. Personally, I never met anybody making this kind of setup so, yeah, I highly doubt it's worth the effort. My stance on that feature is that we should just use a text editor just like it's done in the markdown preview package and not try to mess with styles unless we have data that make it necessary to do so.

mwichary:
Thanks, @abe33. This is very helpful.

I want to push back on the idea this is just about making things “prettier,” though. For some people it’s making things actually _readable,_ and also – since I want to try syntax highlighting elsewhere in the panel – the stakes are higher than they might appear.

I saw enough screenshots of light/dark combinations to make me believe this happens. Besides, it might be a chicken and egg situation – if no plugins work well with this combination, of course no one will use it. But there are some tangible benefits of a mixed contrast combination, as evidenced by for example Slack, which by default uses one.

mwichary:
I’m going to close this since we’ll continue this conversation elsewhere. Thanks again!

PR #377, opened 2018-04-10
Sig caching

dbratz1177:
Does a couple of things:
- Is more selective when making calls to `clientapi/editor/signature` endpoint. Attempts to somewhat mirror what the behavior is on VSCode. The heuristic is based on detecting chars (`,` and `(` that signal the beginning of an argument) and on non-contiguous keystroke events (i.e. the previously cached bufferPosition is not one before the current bufferPosition)
- Changes the insertion behavior of the `signaturePanel` in order to eliminate flashing behavior (currently, the signature panel will toggle on/off with subsequent `loadSignature` actions)

### DIFF
--- a/lib/completions.js
+++ b/lib/completions.js
@@ -6,6 +6,8 @@ const Plan = require('./plan');
 const {delayPromise} = require('./utils');
 let Kite;
 
+let cachedBufferPosition;
+
 const KiteProvider = {
   selector: '.source.python, .source.js',
   disableForSelector: [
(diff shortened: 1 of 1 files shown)


dbratz1177:
@abe33 I think you're right that down the line this kind of logic should be made language specific. I'd imagine that how to do so will become more clear as we add more languages or as more python use cases emerge (i.e. more cases where a quality issue could be solved by very rudimentary parsing done on the plugin side - or maybe so many use cases emerge that we decide to do that kind of work in kited)
I added comments to make sure we had some reminders to revisit when appropriate

PR #389, opened 2018-04-25
Add syntax highlighting to places where it's missing.

mwichary:
@abe33

Adds syntax highlighting so it’s consistently applied in different places in the plugin.

This is the most versatile I could make it – using Highlighter for as much as necessary, and not ever relying on saying “python.” Would welcome any suggestions if you have on how to make it simpler.

Please note that “true” and “false” are not highlighted properly because of this bug: https://github.com/kiteco/kiteco/issues/5616

PS I don’t know why the test is failing. I spent half an hour looking at it, but I’m not sure if it’s a failure or something specific with the test?

Before:
screen shot 2018-04-25 at 13 35 08

After:
screen shot 2018-04-25 at 13 33 30

Before:
screen shot 2018-04-25 at 13 34 58

After:
screen shot 2018-04-25 at 13 33 40

### DIFF
--- a/lib/kite-data-utils.js
+++ b/lib/kite-data-utils.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const {compact, flatten, head, last, uniq, detailGet, detailNotEmpty, detailLang, getFunctionDetails} = require('./utils');
+const {highlightChunk} = require('./highlighter');
 
 const idIsEmpty = (id) =>
   !id || id === '' ||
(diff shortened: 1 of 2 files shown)


lib/kite-data-utils.js
```diff
@@ -12,8 +13,8 @@ const parameterName = (p, prefix = '', w) =>
   p
     ? (
       w
-        ? `<${w}>${prefix}<span class="parameter-name">${p.name}</span></${head(w.split(/\s/g))}>`
-        : `${prefix}<span class="parameter-name">${p.name}</span>`
+        ? `<${w}>${prefix}<span class="syntax--source"><span class="syntax--variable syntax--parameter syntax--function parameter-name">${p.name}</span></span></${head(w.split(/\s/g))}>`
```
abe33: I'm definitely not fond of this, but I guess there's no point for me to argue any more than what I already did.

mwichary:
Oh, wow, thanks for fixing the test. I would’ve never figured that one out!

lib/kite-data-utils.js
mwichary: One thing worth doing might be starting a list of possible future bugs, challenges, and other moments where this requires maintenance? And revisit this solution if it proves to be adding work.

There are changes that improve design that will increase the complexity/messiness of code. Finding a good balance between allowing for some exceptions and craft – and not allowing them to suffocate the code base is part of the job. :·) This feels like the right place to take on some of that debt!

Lastly, I will start a document to enumerate what we’ve done re: syntax highlighting and the challenges, so the scope can be kept in check.

/cc @dbratz1177

PR #391, opened 2018-04-26
Fix status panel settings link

abe33:
There was a missing double quote breaking the link markup.


mwichary:
Argh, that’s on me. Sorry.

PR #394, opened 2018-04-27
Add a visible highlight when the user invokes docs at cursor.

mwichary:
@abe33 @dbratz1177 

Adds an animated highlight for when user invokes Kite via a keyboard shortcut.

Best to play with the branch, rather than for me to show screenshots or animations.

Reasons:
· It’s nice and delightful
· It rewards people using keyboard shortcuts (mastery)
· It highlights the “scope” (region) of what the panel will react to
· It can become Kite’s One Free Interaction (https://www.cooper.com/journal/2009/01/one_free_interaction), something that’s just nice to play with, despite being a no-op
· Functionally, it also turns red if you’re asking for something Kite can’t help with

Unsure if this is the right way to implement it.


mwichary:
@dbratz1177 What are the themes you use?

mwichary:
@abe33 Thanks! I’m not sure what’s the right level of abstraction. Would you mind simplifying it the right way?

abe33:
Sure, let me take that from here

abe33:
@mwichary Done, please look at my comment in the commit for details about the change. Maybe there's a need for small adjustments on the styles as now we're no longer updating the previous decoration.

mwichary:
@abe33 Nice, just tested, this looks exactly the same to me!

@dbratz1177 Last chance to jump in with any colour change suggestions, or you can do it afterwards!

PR #395, opened 2018-04-27
Hide hovers when cursor moves.

mwichary:
@dhung09 + @dbratz1177 

This hides hover pop-up whenever the cursor moves (and/or the programmer starts typing).

The rationale here is that the hover might be in the way as you type, obscuring your view. (The same logic that explains why Mac OS hides a mouse pointer the moment you start typing.)

Since there is nothing in the hover the programmer would want to type _from,_ I feel there is no downside.

Also, should I try to add a test for this?

### DIFF
--- a/lib/kite-editor.js
+++ b/lib/kite-editor.js
@@ -64,6 +64,10 @@ class KiteEditor {
     subs.add(editor.onDidChangeCursorPosition(() => {
       if (!Kite) { Kite = require('./kite'); }
       if (typeof Kite.setStatusLabel === 'function') { Kite.setStatusLabel(this); }
+      if (atom.config.get('kite.enableHoverUI')) {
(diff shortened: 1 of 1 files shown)


dbratz1177:
I like this change and the logic behind it.
As for testing, I think that there ought to be one, but that I don't believe you need to work on putting one in:
* current tests of the hover functionality involve how the hover element loads data of various shapes
* current tests of the kite editor don't have a ton of comprehension currently in terms of covering what is expected to happen in response to user initiated events (as opposed to what it does when receiving data of various types and shapes, for which it does have coverage)
* Adding in support for testing this will likely best involve building infrastructure that extends beyond this particular case
* There's a coming and planned effort to build just that for plugins (esp. VSCode and Atom)

Issue #387, opened 2018-04-24
Standardize commands


dhung09:
Refer to [this doc](https://kite.quip.com/HinNAHAoaVpH/Kite-commands-in-editors) to see the in-depth discussion of these commands.

To summarize, these commands (and only these commands) should be implemented:

- [x] Kite: Docs at cursor
- [x] Kite: Open sidebar
- [x] Kite: Close sidebar
- [x] Kite: Open Error Rescue sidebar
- [x] Kite: Close Error Rescue sidebar
- [x] Kite: Open Copilot
- [x] Kite: Search
- [x] Kite: General settings
- [x] Kite: Permissions
- [x] Kite: Editor plugin settings
- [x] Kite: Status
- [x] Kite: Help
  - This should link to https://help.kite.com/category/43-atom-integration

abe33:
Closing this now, given the PR have been merged

PR #398, opened 2018-05-03
Standardize commands

abe33:
Ref #387


dhung09:
Couple of issues:

- [ ] "Open Sidebar" should bring the docs sidebar to front if there are 2 tabs open in the side panel
- [ ] "Open Error Rescue Sidebar" should bring the Error Rescue sidebar to front if there are 2 tabs open in the side panel
- [ ] "Open Copilot" opens a new tab in the browser. If possible, we should be doing the request within Atom
- [ ] "General Settings" should be opening `kite://settings`
- [ ] "Permissions" should be opening `kite://settings/permissions`

PR #409, opened 2018-05-09
Guard against missing editor of a kite editor in status

abe33:
See https://rollbar.com/Kite/Atom/items/2254/


mwichary:
I just saw this issue, will this solve it?
image

PR #420, opened 2018-05-28
Replace remaining use of localhost to access sidebar settings route

### LINKED ISSUE #416: Blacklist action should use kite:// link
Clicking "Settings..." in this dialog opens a browser which points to http://localhost:46624/settings/permissions?filename=...&action=blacklist 

It should open kite://settings/permissions?filename=...&action=blacklist , though.
(The spec at https://kite.quip.com/4NtPA8CEKNBW/WhitelistingBlacklisting-API#WWKACASpy4a wasn't clear on this. I've fixed this.)

screenshot_20180517_153843

abe33:
Fixes #416


abe33:
@jansorg should be fixed, the test is back as well

PR #431, opened 2018-06-06
Notification queue + editor vacancies notification

abe33:


lib/kite.js
```diff
@@ -328,6 +313,125 @@ const Kite = module.exports = {
...
+        () => !atom.inSpecMode()),
+    });
+
+    this.notificationQueue.add(() => {
+      if (!DataLoader) { DataLoader = require('./data-loader'); }
+
+      if (!atom.config.get('kite.showEditorVacanciesNotificationOnStartup')) {
+        return null;
+      }
+      metrics.sendFeatureMetric('atom_editor_vacancy_notification_requested');
```
dhung09: can't we use `metrics.featureRequested('editor_vacancy_notification')` instead?

lib/kite.js
```diff
@@ -328,6 +313,125 @@ const Kite = module.exports = {
...
+
+      return DataLoader.getPlugins()
+      .then(({plugins}) => {
+        const vacancies = plugins
+        .filter(p => !p.encountered && p.editors.length && !p.editors.some(e => e.plugin_installed));
+
+        const buttons = [{
+          text: 'Show me',
+          onDidClick(dismiss) {
+            metrics.sendFeatureMetric('atom_editor_vacancy_notification_install_requested');
```
dhung09: can't we use `metrics.featureRequested('editor_vacancy_notification_install')` instead?

lib/kite.js
```diff
@@ -328,6 +313,125 @@ const Kite = module.exports = {
...
+            atom.applicationDelegate.openExternal('kite://settings/plugins');
+            metrics.sendFeatureMetric('atom_editor_vacancy_notification_install_fulfilled');
```
dhung09: can't we use `metrics.featureFulfilled('editor_vacancy_notification_install')` instead?

lib/kite.js
```diff
@@ -328,6 +313,125 @@ const Kite = module.exports = {
...
+            dismiss();
+          },
+        }, {
+          text: 'Never ask again',
+          onDidClick(dismiss) {
+            metrics.sendFeatureMetric('atom_editor_vacancy_notification_disabled_requested');
```
dhung09: can't we use `metrics.featureRequested('editor_vacancy_notification_disable')` instead?

lib/kite.js
```diff
@@ -328,6 +313,125 @@ const Kite = module.exports = {
...
+            atom.config.set('kite.showEditorVacanciesNotificationOnStartup', false);
+            metrics.sendFeatureMetric('atom_editor_vacancy_notification_disabled_fulfilled');
```
dhung09: can't we use `metrics.featureFulfilled('editor_vacancy_notification_disable')` instead?

lib/kite.js
```diff
@@ -328,6 +313,125 @@ const Kite = module.exports = {
...
+            dismiss();
+          },
+        }];
+
+        // If we have some editors in vacancies we'll display the notification as
+        // there's no other conditions
+        metrics.sendFeatureMetric('atom_editor_vacancy_notification_fulfilled');
```
dhung09: can't we use `metrics.featureFulfilled('editor_vacancy_notification')` instead?

lib/kite.js
```diff
@@ -328,6 +313,125 @@ const Kite = module.exports = {
...
+          text: 'No',
+          onDidClick(modal) {
+            atom.config.set('kite.editorMetricsEnabled', 'no');
+            modal.destroy();
+          },
+        },
+      ],
+    }, {
+      condition: fromSetting('kite.editorMetricsEnabled', 'not specified'),
+      onDidDismiss: res => {
```
dhung09: i believe this belongs in above, along with all the other editor vacancy logic

lib/kite.js
```diff
@@ -328,6 +313,125 @@ const Kite = module.exports = {
...
+        // res is the text of the button when it's an action that leads
+        // to the notification dismissal
+        if (!res) {
+          metrics.sendFeatureMetric('atom_editor_vacancy_notification_dismiss_requested');
```
dhung09: see comments above about using `metrics.featureRequested` and `metrics.featureFulfilled`

lib/kite.js
abe33: oh, yeah totally, it's just that recently I was thinking of removing these two functions, given we've introduced new metrics that does not conform to that sheme (`atom_*_(requested|fulfilled)`) but yeah for now I can make the change

dhung09: i think it makes sense to keep them! if i had known about those convenience functions earlier, i would have named those counters differently to start with 🙂

PR #432, opened 2018-06-06
Switch notification center to use a queue for all notifications

abe33:
Includes moving the kite queue and startup notification to the center
and implement events from notification to queue to center.

### DIFF
--- a/lib/kite-modal.js
+++ b/lib/kite-modal.js
@@ -1,10 +1,33 @@
 'use strict';
 
+const {Emitter} = require('atom');
+
 module.exports = class KiteModal {
   constructor({content, buttons} = {}, options = {}) {
     this.content = content;
     this.buttons = buttons;
     this.options = options;
+    this.emitter = new Emitter();
+  }
+
+  onDidNotify(listener) {
+    return this.emitter.on('did-notify', listener);
+  }
+
+  onDidRejectNotification(listener) {
+    return this.emitter.on('did-reject-notification', listener);
+  }
+
+  onDidDismissNotification(listener) {
+    return this.emitter.on('did-dismiss-notification', listener);
+  }
+
+  onDidClickNotificationButton(listener) {
+    return this.emitter.on('did-click-notification-button', listener);
+  }
+
+  dismiss() {
+    this.destroy();
   }
 
   destroy() {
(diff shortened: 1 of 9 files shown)


dhung09:
It seems like the whitelist notifications are being queued when they shouldn't be. To reproduce:

1. Have `showWelcomeNotification` be `true`
2. Open Atom, let the welcome notification render and stay open
3. Open an un-whitelisted file
4. Open to a whitelisted file
5. Close the welcome notification

After the welcome notification is closed, a whitelist notification will be shown, even though the currently opened file is whitelisted.

The expected behavior is that no whitelist notification should be shown. At the time the un-whitelisted file was focused, the welcome notification was still being shown, so the whitelist notification did not need to have been shown.

If the user had closed the welcome notification while the un-whitelisted file was open, I think it's OK if the whitelist notification isn't shown immediately afterwards. We could wait for the next user event to trigger a 403 before showing the notification.

abe33:
@dhung09 Thanks for finding this one! I've pushed a fix for it, basically the whitelist notification now has an extra condition which is that the active file must be the same than the one for which we want to display the whitelist, if that's not the case we won't show anything, and as we didn't show the notification coming back to that file will queue a notification again.

dhung09:
One more thing I noticed:

1. Open Atom. Dismiss the welcome notification if it's shown.
2. Open an un-whitelisted file `foo.py`. The whitelist notification for `foo.py` shows up as expected.
3. Open another un-whitelisted file `bar.py`. The whitelist notification for `bar.py` replaces the current notification as expected.
4. Switch back to `foo.py`

I would expect the notification for `foo.py` to be shown, but instead, the notification for `bar.py` is shown

abe33:
@dhung09 latest commit fixes that and adds a test for that case as well

dhung09:
@abe33 now I have an issue where 2 notifications from Kite will show at the same time:

1. Make sure `showWelcomeNotification` is set to `true`
2. Open Atom and keep the welcome notification open
3. Open an unwhitelisted file

Both the welcome notification and the whitelist notification will be shown. In this case, the welcome notification should take precedence over the whitelist notification

abe33:
@dhung09 I don't reproduce on this branch, is it possible that your local version on this branch was overriden by master during a kite update (this is quite frequent and I can get fooled by that myself). As master now has the notification queue, the behaviour of the first notifications are still ok, just all the other notifications don't go through the queue.

Issue #434, opened 2018-06-08
Kite status icon flickers


dhung09:
I've noticed that the Kite icon in the status bar flickers constantly

![image](https://user-images.githubusercontent.com/3752225/41173098-9b335612-6b0a-11e8-9e9f-7e0813f0838b.png)

Here it's empty, but it'll come back intermittently as I click around a Python file

dhung09:
@abe33 this is still happening for me as of plugin v0.97.0

Issue #430, opened 2018-06-05
Permissions in status panel not disabled when Kite isn't running


jansorg:
I've noticed this in branch cn-replace-kite-installer-use. I didn't check if this is in master, too.
Clicking "Permissions" opens a link in the web browser, Kite isn't running though. It should be disabled.

![bildschirmfoto 2018-06-05 um 14 52 46](https://user-images.githubusercontent.com/11473063/40977086-40c28b3a-68d0-11e8-8166-75ec76ec5ce9.png)

abe33:
Right, this is also happening on master, the link lack a `account-dependent` class to make it behaves just like the `Kite Settings` action above.

PR #436, opened 2018-06-11
Fix various issues

abe33:
Addresses #430, #433 and #434


jansorg:
Looks good on Windows. I'll do some more testing on macOS

PR #443, opened 2018-06-18
JSON tests generation experiment

abe33:


spec/json/tests/sample.json
```diff
@@ -0,0 +1,76 @@
...
+      }
+    },
+    {
+      "description": "sends edit event",
+      "step": "expect",
+      "type": "request",
+      "properties": {
+        "path": "/clientapi/editor/event",
+        "method": "POST",
+        "body": {
```
jansorg: @abe33 Do you prefer a filename (as above) or inlined JSON (as shown here)?

jansorg:
@abe33 Sorry, I approved too quickly because  I assumed that this PR only contains changes to tests.

PR #465, opened 2018-07-25
Remove custom signature detection rules

### LINKED ISSUE #415: signatures request not always made inside call expression
out

In particular, when typing the final close paren in `map(filter(123)`, a request is not made. The modal for the `filter` is sometimes, but not always, closed:

```
[kited] 2018/05/16 15:42:35.234954 manager.go:130: got valid signatures request for /Users/naman/Desktop/kitetest/foo/bar/__init__.py:4
[kited] 2018/05/16 15:42:35.251944 handlers.go:64: POST /clientapi/editor/signatures 200 6088 17.158127ms
[kited] 2018/05/16 15:42:38.943215 manager.go:130: got valid signatures request for /Users/naman/Desktop/kitetest/foo/bar/__init__.py:10
[kited] 2018/05/16 15:42:38.943726 handlers.go:64: POST /clientapi/editor/signatures 200 6088 668.755µs
[kited] 2018/05/16 15:42:39.777498 manager.go:130: got valid signatures request for /Users/naman/Desktop/kitetest/foo/bar/__init__.py:11
[kited] 2018/05/16 15:42:39.793072 handlers.go:64: POST /clientapi/editor/signatures 200 3059 15.708119ms
[kited] 2018/05/16 15:42:49.134484 manager.go:130: got valid signatures request for /Users/naman/Desktop/kitetest/foo/bar/__init__.py:15
[kited] 2018/05/16 15:42:49.151703 handlers.go:64: POST /clientapi/editor/signatures 200 6088 17.346739ms
[kited] 2018/05/16 15:42:49.538421 manager.go:130: got valid signatures request for /Users/naman/Desktop/kitetest/foo/bar/__init__.py:14
[kited] 2018/05/16 15:42:49.558255 handlers.go:64: POST /clientapi/editor/signatures 200 3059 19.963534ms
[kited] 2018/05/16 15:42:51.807103 manager.go:130: got valid signatures request for /Users/naman/Desktop/kitetest/foo/bar/__init__.py:14
[kited] 2018/05/16 15:42:51.807545 handlers.go:64: POST /clientapi/editor/signatures 200 3059 581.824µs
[kited] 2018/05/16 15:42:53.790275 manager.go:130: got valid signatures request for /Users/naman/Desktop/kitetest/foo/bar/__init__.py:16
[kited] 2018/05/16 15:42:53.805801 handlers.go:64: POST /clientapi/editor/signatures 400 92 15.689702ms
[kited] 2018/05/16 15:42:54.155393 manager.go:130: got valid signatures request for /Users/naman/Desktop/kitetest/foo/bar/__init__.py:15
[kited] 2018/05/16 15:42:54.172603 handlers.go:64: POST /clientapi/editor/signatures 200 6088 17.376379ms
[kited] 2018/05/16 15:42:54.544853 manager.go:130: got valid signatures request for /Users/naman/Desktop/kitetest/foo/bar/__init__.py:14
[kited] 2018/05/16 15:42:54.560973 handlers.go:64: POST /clientapi/editor/signatures 200 3059 16.249017ms
```

abe33:
They were messing up the whole check of being inside a signature.

Closes #415

### DIFF
--- a/lib/completions.js
+++ b/lib/completions.js
@@ -61,11 +61,12 @@ const KiteProvider = {
     // Prior to version 1.24 the arguments scope didn't had the function-call|method-call
     // prefix, so we need to keed the old check otherwise we're breaking the signature
     // panel in these versions
-    if (parseFloat(atom.getVersion()) <= 1.23) {
-      return scopeDescriptor.scopes.some(s => /(function|method)-call|arguments/.test(s));
-    } else {
-      return scopeDescriptor.scopes.some(s => /(function|method)-call\.arguments/.test(s));
-    }
+    return (parseFloat(atom.getVersion()) <= 1.23
+      ? scopeDescriptor.scopes.some(s => /(function|method)-call|arguments/.test(s))
+      : scopeDescriptor.scopes.some(s => /(function|method)-call\.arguments/.test(s)))
+      // the opening parenthesis in an outer call doesn't have the function-call scope
+      // but has an arguments.bracket one
+      || scopeDescriptor.scopes.some(s => /arguments\.bracket/.test(s));
   },
 
   loadSignature({editor}) {
--- a/lib/completions.js
+++ b/lib/completions.js
@@ -33,7 +33,7 @@ const KiteProvider = {
           [position.row, position.column - 1],
           [position.row, position.column],
         ]);
-        if (this.isInsideFunctionCall(params) && /[\(,=]$/.test(prefixChar)) {
+        if (this.isInsideFunctionCall(params) && /[\(\),=\s]$/.test(prefixChar)) {
           promise = promise.then(() => this.loadSignature(params));
         } else {
           this.clearSignature();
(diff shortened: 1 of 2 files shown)


lib/completions.js
```diff
@@ -62,20 +62,9 @@ const KiteProvider = {
...
     } else {
-      return scopeDescriptor.scopes.some(s =>
-        s.indexOf('function-call') !== -1 ||
-        s.indexOf('method-call') !== -1
-      ) &&
-      scopeDescriptor.scopes.some(s =>
-        s.indexOf('punctuation.definition.arguments.end.bracket.round.python') !== -1 ||
-        /(function|method)-call\.arguments/.test(s)
-      );
+      return scopeDescriptor.scopes.some(s => /(function|method)-call\.arguments/.test(s));
```
dbratz1177: Probably `|` as opposed to `\.` in the regex here? If I make the switch, I get correct behavior

abe33: I think we definitely need to check for both function/method call and arguments at the same time (as opposed with the >= 1.23 variant where this scope didn't existed)

Here's an example: 

here the `function-call` scope wraps the entire expression: 
sample_py_ ___github_kite

but `function-call.arguments` will match only that part: 
sample_py_ ___github_kite

which is what we want based on the specs (no signature request in the function name, but it should start with the opening parenthesis), then as long as we are in an arguments section we can trigger signature requests based on the previous character (using the `/\(\),=\s/` expression)

Finally, when you finally end outer top call the closing parenthesis doesn't belong to the arguments section, and if we were using only `function-call` or `arguments`,  and given we want `)` to trigger signature requests (when we end an inner function call and want to update the signature panel to now show the outer function call) we would endup trigger a signature request for the `)` of the outer function, which is not what the spec describe. 
sample_py_ ___github_kite

dbratz1177: Your logic and reading of the spec seems sound. However, in tests, I still notice that while the layered call situation works well:
screen shot 2018-08-07 at 9 23 51 am

The single function/method call does not open up the signatures panel:
screen shot 2018-08-07 at 9 24 27 am

abe33: Oh you're right, pushing a fix.

abe33: This case is currently not covered by the json tests (the other use case were and the tests were not passing, hence the change), I'll add one as soon as we release the feature system I started working on to support this kind of case (auto bracket/parentheses closing) that are not always available in an editor

dbratz1177:
@abe33 still not working for me with the outer function call 🙁 
Here's what the `ScopeDescriptors` look like for `json.dumps()` and `json.dumps(filter())`, respectively, with the cursor in both cases being inside `()`:
screen shot 2018-08-08 at 10 10 46 am

PR #481, opened 2018-08-27
Bump kite-installer, kite-api and kite-connector

abe33:
This introduces the new error types for install flow error reporting.


abe33:
@dbratz1177 FWIW I went with a major bump this time because the change in error types will breaks code that relied on the previous version.

PR #485, opened 2018-09-06
Allow , and ' ' to trigger signature with cursor in closing paren

### LINKED ISSUE #484: signatures request not made when matching close paren exists after cursor
e.g. if I type

```
import json
json.dumps(|)
```

then the signature request is made and the panel pops up. However, after typing 

```
import json
json.dumps(x, |)
```

the signature panel does not reappear, and a request is not remade to kited.

On the other hand, if the closing paren does not exist (i.e. I manually delete it, since Atom creates the close paren automatically), the signature request is remade and the panel pops up.
```
import json
json.dumps(x, |
```

cc @dbratz1177

abe33:
Fixes #484

### DIFF
--- a/lib/completions.js
+++ b/lib/completions.js
@@ -86,7 +86,7 @@ const KiteProvider = {
       // method calls
       (
         scopeDescriptor.scopes.some(s => /arguments(.*)\.bracket/.test(s)) &&
-        /[\(]$/.test(prefixChar)
+        /[\(, ]$/.test(prefixChar)
       );
   },
 


dhung09:
@abe33 I'm not sure if it's explicitly the `,` and `''` chars that are preventing the signatures request from triggering. when both the opening and closing parenthesis are present, it seems _any_ character doesn't trigger the request:

both-parens

When only the opening parenthesis present, it seems like the `,` triggers requests, but other characters don't:

open-parens

adamsmith:
Can we add an automated regression test for this?

abe33:
@dhung09 Oh yeah right, the spec changed on that, I was still on the old version where signature requests were not supposed to occurs when typing an argument/value.

@adamsmith @jansorg has been working on the news json tests for these specs so it should land soon in the common tests suite

PR #490, opened 2018-09-13
New signature specs support

abe33:
Allow request on any characters typed in a function call as well as updating the signature panel on cursor move inside a function call when the panel is visible.

### DIFF
--- a/lib/completions.js
+++ b/lib/completions.js
@@ -29,11 +29,8 @@ const KiteProvider = {
       ? delayPromise(() => {
         let promise = Plan.queryPlan();
         const position = params.editor.getCursorBufferPosition();
-        const prefixChar = params.editor.getTextInBufferRange([
-          [position.row, position.column - 1],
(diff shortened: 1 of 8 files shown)


lib/completions.js
```diff
@@ -29,14 +29,18 @@ const KiteProvider = {
...
-        if (this.isInsideFunctionCall(params, prefixChar)) {
+        if (this.isInsideFunctionCall(params.editor, position) ||
+            this.isInsideFunctionCallBrackets(params.editor, position)) {
           promise = promise.then(() => this.loadSignature(params));
         } else {
-          this.clearSignature();
+          // the signature panel is visible, it's either we've deleted the
+          // open paren or got past the closing one. We do an extra request
+          // to conform to the specs
+          if (this.signaturePanel && this.signaturePanel.parentNode) {
```
dbratz1177: tiny nit: use `isSignaturePanelVisible()`?

PR #491, opened 2018-09-17
Experiment separated signature

abe33:
This branch is based off #490 to benefit from the new specs tests. Surprisingly, making the new system passes the tests was way easier than the old one as we can control finely when to trigger requests.


abe33:
When two function calls are lined up vertically so that moving the cursor up or down places it still in a function call (but a different one)

1. ```
foo(1|, 3)
bar(2, 4)
```

2. ```
foo(1, 3)
bar(2|, 4)
```

Shall we keep the panel open? Shall we close it? If we want to close the panel, how do we know the function is a different one? In atom I'm relying on how the grammar highlights the code so there're ambiguities.

This one concern the new panel. When the panel is opened while the cursor is in the bottom line, the scope range we get from Atom's API only spans that last line. That means the panel will be overlapping the top line.

```
foo(1,
    3|)
```

Only solution here seems to walk back from the cursor to find the call start or rely on new data from kited to place the panel.

dhung09:
@abe33 

> When two function calls are lined up vertically so that moving the cursor up or down places it still in a function call (but a different one)

In this case, I think it's ok to have the select event that moves the cursor to the other function call still trigger a signatures request. It's not ideal, but it seems like this is the easiest approach to deal with an edge case. Also, if the behavior was unintended, it should be easy enough for the user to press `Esc` to close the panel.

PR #499, opened 2018-09-26
Hide status bar content and clear tooltip when no active supported file

abe33:


jansorg:
@abe33 Works well here

PR #502, opened 2018-09-27
Remove login UI and open copilot instead

abe33:
All places were already using a single function in the app object.


jansorg:
@abe33 Successfully tested on Windows 10. 
Is there a reason why `kite://open` was used? I only know about `kite://home` so far.

abe33:
No reason, just silly mistake given I tried that first in terminal, worked so I moved on. This is fixed.

BTW, CI is now failing cause Github rolled out new beta with tree-sitter enabled by default, breaking our signatures tests.

abe33:
Yes tests are fine, the failures comes from the lack of scopes in editors (due to tree-sitter and the new grammar not handling functions atm), not from the login-related tests. I'll probably disable CI on beta until I found out how they plan to update language python grammar.

PR #503, opened 2018-10-02
Make new completions specs pass

abe33:


package.json
```diff
@@ -128,6 +128,6 @@
     "fbjs": "^0.8.6",
     "javascript-obfuscator": "^0.8.3",
     "sinon": "^2.3.5",
-    "editors-json-tests": "git://github.com/kiteco/editors-json-tests.git#master"
+    "editors-json-tests": "git://github.com/kiteco/editors-json-tests.git#new-completion-spec"
   }
```
jansorg: Revert this change because the json tests are in master now?

abe33: oh yeah right

PR #516, opened 2018-10-31
Implement tree-sitter warning notification

### LINKED ISSUE #514: Offer to turn off tree-sitter parsing on startup
(See #512 for context)

On plugin startup, we should detect if the user is using the tree-sitter parser and offer to turn it off.

You can detect this setting via:
```javascript
atom.config.get('core.useTreeSitterParsers')
```

The notification to be shown should look like:
image

Code to generate the notification (we'll replace the link with another one):
```javascript
atom.notifications.addWarning('Kite is not configured properly', {
  description: 'You must turn off Atom\'s tree-sitter parser for Kite to show you information on how to call functions. <a href="https://kite.com">Learn more</a>',
  dismissable: true, 
  buttons: [{text: 'Turn off'}]
})
```

Clicking on the "Turn off" button should disable the tree-sitter setting.

This notification should be shown on _every_ startup because the tree-sitter setting takes away a signification portion of Kite's functionality.

abe33:
Closes #514

### DIFF
--- a/lib/kite.js
+++ b/lib/kite.js
@@ -230,7 +230,7 @@ const Kite = module.exports = {
     this.subscriptions.add(atom.config.observe('core.useTreeSitterParsers', (using) => {
       if (using) {
         const notif = atom.notifications.addWarning('Kite is not configured properly', {
-          description: 'You must turn off Atom\'s tree-sitter parser for Kite to show you information on how to call functions. <a href="https://kite.com">Learn more</a>',
+          description: 'You must turn off Atom\'s tree-sitter parser for Kite to show you information on how to call functions. <a href="https://help.kite.com/article/80-why-do-i-need-to-turn-off-tree-sitter">Learn more</a>',
           dismissable: true,
           buttons: [{
             text: 'Turn off',
--- a/lib/kite.js
+++ b/lib/kite.js
@@ -227,6 +227,23 @@ const Kite = module.exports = {
       Logger.LEVEL = Logger.LEVELS[level.toUpperCase()];
     }));
 
+    this.subscriptions.add(atom.config.observe('core.useTreeSitterParsers', (using) => {
+      if (using) {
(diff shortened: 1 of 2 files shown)


lib/kite.js
```diff
@@ -227,6 +227,23 @@ const Kite = module.exports = {
...
 
+    this.subscriptions.add(atom.config.observe('core.useTreeSitterParsers', (using) => {
+      if (using) {
+        const notif = atom.notifications.addWarning('Kite is not configured properly', {
+          description: 'You must turn off Atom\'s tree-sitter parser for Kite to show you information on how to call functions. <a href="https://kite.com">Learn more</a>',
+          dismissable: true,
+          buttons: [{
+            text: 'Turn off',
+            onDidClick() {
+              console.log('here');
```
dbratz1177: take out log

dhung09: We also need to restart Atom for the setting change to kick in. We can change the button text to "Turn off and restart"

dbratz1177: will we want to programmatically restart atom? Is that even possible with their API? If it isn't, we'll have to adjust the copy such that it's clear to the user that they'll have to take action in that way

dhung09: Updated "Learn more" link: https://help.kite.com/article/80-why-do-i-need-to-turn-off-tree-sitter

abe33: > We also need to restart Atom for the setting change to kick in. We can change the button text to "Turn off and restart" 

That's definitely not what I'm seeing: 
tree-sitter-update

dbratz1177: Yeah, I can confirm you don't need a restart to see the effects

dhung09: This may be dependent on what version of Atom you're running. On v1.32.0 you need to restart:

oct-31-2018 11-28-38

So I've added in an `atom.reload()` in there as well

PR #525, opened 2018-11-12
Make signature separated but still sticking to completions

abe33:


lib/kite.js
```diff
@@ -554,23 +561,30 @@ const Kite = module.exports = {
     atom.packages.activatePackage('autocomplete-plus')
     .then(autocompletePlus => {
       const acp = autocompletePlus.mainModule;
-      const manager = acp.autocompleteManager || acp.getAutocompleteManager();
+      const manager = acp.autocompleteManager;
+
+      if (!manager) {
+        console.log('no manager');
```
dbratz1177: log

abe33:
@dbratz1177 Addressed the first two points, for the last one it's actually much complex than just making it scrollable, we never set a fixed height and also we're fixing the panel width which that our previous measurement might be now off (we'll have to measure again, forcing a browser layout, to check when the panel goes past the window boundaries). Also, I tried a naive approach using `max-height` and there's a lot of flicker while scrolling as it seems the panel height is kept on our next evaluation and then reset the one after that. Not sure what strategy to use atm.

PR #533, opened 2018-11-26
Implement support for kited-test

abe33:
Includes changing the `request` and `request_count` expectation to hit the request history endpoint and enabling the live testing using a `LIVE_ENVIRONMENT` variable.

I had to deactivate an unrelated test to get the build pass. I'll look into this later.


jansorg:
@abe33 I've updated to match an update to the url of kited-test

PR #536, opened 2018-11-28
Implement kited autostart

### LINKED ISSUE #534: Autostart kited at startup
Autostart kited at startup of the plugin.

- if necessary add a new setting to disable the automatic startup, default to false, i.e. to an automatic start of kited
- if setting "start kited at startup" is enabled:
    - detect if kited is running at startup
         - if not running: detect if kited is installed
            - if installed: **launch kited**

abe33:
Closes #534

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,11 @@
       "default": false,
       "title": "Developer mode",
       "description": "Enable Kite developer mode, displaying received data in the various UIs"
+    },
+    "startKitedAtStartup": {
+      "type": "boolean",
+      "default": false,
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -86,6 +86,11 @@
       "default": false,
       "title": "Developer mode",
       "description": "Enable Kite developer mode, displaying received data in the various UIs"
+    },
+    "startKitedAtStartup": {
+      "type": "boolean",
+      "default": false,
```
jansorg: Should default to `true`

PR #537, opened 2018-11-28
Change status bar to conform with new specs

### LINKED ISSUE #535: Update to status bar
https://kite.quip.com/rVcJA1bIqZny/Kite-Status-Spec#QcWACAl5h2O contains an update to the status bar.
- implement changes to match the updated table
- remove the "docs available" message from the statusbar

abe33:
Closes #535


jansorg:
@abe33 Hm, I'm not really seeing changes here. I'm not sure if this is my setup.

- The icon still stays visible for a few seconds when the last supported file was closed.  Is it possible to hide it immediately?
- When the icon is hidden, then there's a larger bit of blank space where the icon was before. Is it possible to remove this when the icon isn't visible? This might look odd if other plugins add more content to the left of kite.
screenshot_20181205_155001

PR #544, opened 2018-12-06
Refactor status as a module

abe33:
The main changes here are that we're no longer use a custom element for the status bar view and that all checks and rendering are now done in one place.
Also, checking Kite's status is now much more simpler, if we're in a supported file we only contact the `/clientapi/status` endpoint and infer Kite's state from there using the new errors from `kite-connector`. In the case of unsupported/no files we're simply not making any request and clean up the status.


abe33:
@dbratz1177 
1. This is weird, the sentence `Kite is not enabled for the current directory` is no longer present in the whole project, so I wonder how it could be displayed there.
2. Same thing as above, `unknown state` is not present in the project except for one test description, also, the only write access to the `tooltipText` property is done inside `status.js` file and only using either a string from the `TOOLTIPS` object or an empty string.

PR #545, opened 2018-12-10
Remove suggestions limit along with signature

abe33:
Remove the setting and the logic in the signature panel that was changing the suggestion list value at runtime when the signature was visible.


dbratz1177:
Is this still needed with #547?

abe33:
It depends, do we want to keep that or not, if yes, then we can close that PR.

PR #573, opened 2019-02-12
Hover instance kind

### LINKED ISSUE #385: hover + completions interaction
surfaced during QA, when completions are triggered simultaneously to hovering over a token:
image

Not sure to what degree this is fixable or if this is even something to consider a problem, but wanted to get thoughts 

@abe33 @mwichary

dbratz1177:
addresses 3 of the issues mentioned [here](https://kite.quip.com/aSWPAxHqAe9o) (hover instance type, hover over signature and completions panel, status panel hiding), and additionally fixes some lingering test failures

fixes #385

### DIFF
--- a/lib/elements/kite-hover.js
+++ b/lib/elements/kite-hover.js
@@ -2,7 +2,7 @@
 
 require('./kite-links');
 const {Point, CompositeDisposable} = require('atom');
-const {symbolName, symbolKind} = require('../kite-data-utils');
+const {symbolName, symbolType} = require('../kite-data-utils');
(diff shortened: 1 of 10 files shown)


its-dhung:
I realize this isn't spec'ed anywhere but we should limit the hint to the first 3 types only. Otherwise the description becomes too long.

image

This should be rare, but we should account for this case. With regards to the ordering of types, ideally the backend would return the types in some sorted order (e.g. highest probability type to lowest probability), but it shouldn't be the plugin's job to determine this.

PR #578, opened 2019-03-05
Minimize screen real estate

dbratz1177:
Addresses the remaining issues described here: https://kite.quip.com/aSWPAxHqAe9o
General efforts were applied to making the Atom UI less intrusive, more space efficient, more consistent across elements, and more parseable visually

More specifically:
- padding, margins, and font-sizes were reduced across signature, completions, and hover elements 
- the hover UI is now restricted to a single line
- unicode icons were introduced to aid with the parsing of what certain actions do
- some code was opportunistically reorganized
- return types were added to signatures
- popular patterns are now collapsable (ala Sublime)

### DIFF
--- a/lib/data-loader.js
+++ b/lib/data-loader.js
@@ -69,7 +69,7 @@ const DataLoader = {
         };
         if (c.symbol) {
           completion.type = symbolTypeString(c.symbol);
-          completion.rightLabelHTML = completionRightLabel(c.symbol);
+          completion.rightLabelHTML = kindLabel(c.hint);
         } else {
           completion.type = c.hint;
           completion.rightLabelHTML = kindLabel(c.hint);
--- a/lib/data-loader.js
+++ b/lib/data-loader.js
@@ -62,15 +62,17 @@ const DataLoader = {
     return KiteAPI.getCompletionsAtPosition(editor.getPath(), text, cursorPosition, 'atom')
     .then(completions => {
       return completions.map((c) => {
-        console.log('c.hint', c.hint, c);
-        return {
+        const completion = {
           text: c.insert || c.display,
           displayText: c.display,
           type: symbolTypeString(c.symbol),
           rightLabelHTML: completionRightLabel(c.symbol),
-          description: c.documentation_text || 'kite',
-          descriptionHTML: appendCompletionsFooter(c, escapeHTML(c.documentation_text)),
         };
+        if (c.documentation_text) {
+          completion.descriptionHTML = appendCompletionsFooter(c, escapeHTML(c.documentation_text));
+          completion.description = c.documentation_text;
+        }
+        return completion;
       });
     });
   },
--- a/lib/data-loader.js
+++ b/lib/data-loader.js
@@ -1,10 +1,10 @@
 'use strict';
 
-const {internalExpandURL} = require('./urls');
 const KiteAPI = require('kite-api');
 const KiteConnect = require('kite-connector');
 const metrics = require('./metrics');
 const {escapeHTML} = require('./highlighter');
+const {appendCompletionsFooter, symbolTypeString, completionRightLabel} = require('./kite-data-utils');
 
 const pendingStatusRequestByEditor = {};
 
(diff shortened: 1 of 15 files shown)


dbratz1177:
comment per @tarakju - have plugin remember whether `Patterns` was expanded by a user or not

styles/kite-signature.less
```diff
@@ -3,29 +3,49 @@
...
+    }
   }
 
-  pre {
+  pre, code {
     padding: 0;
     background: transparent;
     cursor: default;
     padding-left: 4ch;
     text-indent: -4ch;
```
its-dhung: I would suggest changing these values to `-2ch` to save horizontal space. We'll also need to change the indent for the popular patterns.

lib/data-loader.js
```diff
@@ -62,38 +62,27 @@ const DataLoader = {
...
           text: c.insert || c.display,
           displayText: c.display,
-          type: c.hint,
-          rightLabel: c.hint,
-          description: c.documentation_text || 'kite',
-          descriptionHTML: this.appendCompletionsFooter(c, escapeHTML(c.documentation_text)),
+          className: 'kite-completion',
         };
+        if (c.symbol) {
+          completion.type = symbolTypeString(c.symbol);
```
its-dhung: I would suggest utilizing `c.hint` here, even if there is a symbol attached to the completion. This allows the backend to annotate hints using extra information that the plugin does not know about, which can then still be rendered by the plugin. This behavior becomes more important as we start adding more and more intelligence into our completions since we can use hints to explain to the user why certain completions show up.

package.json
```diff
@@ -71,6 +71,18 @@
...
+    "signatureKwargsVisible": {
+      "type": "boolean",
+      "default": false,
+      "title": "Kite Signature panel **KW section",
+      "description": "Toggle visibility for the **KW section in the Kite Signature Panel"
+    },
+    "signaturePopularPatternsVisible": {
+      "type": "boolean",
+      "default": false,
+      "title": "Kite Signature panel Examples section",
```
its-dhung: Change this to:
```
"title": "Show function call examples",
"description": "Show examples on how to call a function when the function signature panel is shown"
```

package.json
```diff
@@ -71,6 +71,18 @@
       "default": true,
       "title": "Start Kite Engine on startup",
       "description": "Automatically start Kite Engine on editor startup if it's not already running."
+    },
+    "signatureKwargsVisible": {
+      "type": "boolean",
+      "default": false,
+      "title": "Kite Signature panel **KW section",
```
its-dhung: Change this to:
```
"title": "Show function keyword arguments",
"description": "Show inferred keyword arguments for a function when the function signature panel is shown"
```

styles/kite-signature.less
```diff
@@ -3,29 +3,49 @@
...
+    color: @text-color-subtle;
+    padding: 0 0.1em;
+  }
+
   h4 {
     font-size: 0.9em;
-    margin: 15px 0 2px;
+    margin: 10px 0 2px;
+
+    .popular-patterns-toggle, .kwargs-toggle {
```
its-dhung: Not related to this code specifically, but it seems like the type hints for inferred keyword arguments are rendered in a larger font than the argument names:
image
We should make them the same size.

lib/data-loader.js
dbratz1177: Makes sense. For the `type` field, though, I think it's more appropriate to use information from the symbol. autocomplete-plus uses the `type` to derive the lefthand icon, and so expects bare bones strings. Based on what you said, I think we should use `hint` for `rightLabelHTML`

styles/kite-completions.less
```diff
@@ -0,0 +1,114 @@
...
+          &.selected {}
+
+          .icon-container {
+            i.icon {
+              position: relative;
+
+              &::after {
+                content: '';
+                display: block;
+                width: 0.78em;
```
its-dhung: I tried some different values for this, and I think `0.9em` for both width and height is acceptable. It's large enough to be definitely noticeable, but not too large that it blocks the native icon.

PR #588, opened 2019-04-17
bundling for atom

dbratz1177:
Adds bundling for atom using webpack. Only worrying about JS files, as Atom has its own way of loading styles, and, in our JS code, we get image assets (and compute their paths) at runtime

Due to how atom packages are published / configured, we don't `.gitignore` the bundled file (in this case, `dist/main.js`), nor is there an equivalent of `.vscodeignore` to ignore the source files in the construction of this package. That's ok as it's relatively small file size wise, and the major performance gains are from better load and activation times:

Pre-bundling:
Screen Shot 2019-04-17 at 11 16 56 AM

Post-bundling:
Screen Shot 2019-04-17 at 1 12 22 PM

As a follow up to this, it may be wise to reevaluate our lazy-loading module strategy, which currently shunts startup costs to activation vs. load

sibling pr for vscode: https://github.com/kiteco/vscode-plugin/pull/224

cc @dhung


dbratz1177:
merge with https://github.com/kiteco/kiteco/pull/7810

9994444ggg:
Tank you

9994444ggg:
Thank You helps me very good

9994444ggg:
Thank You Helps me Good

edzkite:
Checked out the package locally. When I run `npm install`, this is the error:

```
> kite@0.154.0 prepublish /Users/edwardzhao/go/src/github.com/kiteco/atom-plugin
> rm -rf node_modules && rm package-lock.json && apm install && rm -rf dist && npm run build-prod

rm: package-lock.json: No such file or directory
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! kite@0.154.0 prepublish: `rm -rf node_modules && rm package-lock.json && apm install && rm -rf dist && npm run build-prod`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the kite@0.154.0 prepublish script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm WARN Local package.json exists, but node_modules missing, did you mean to install?
```

Looks like as part of the `prepublish` step, it tries to remove `package-lock.json`, but since it doesn't exist in the workspace yet, it fails.

tarakju:
@edzkite I think this can be resolved by updating the prepublish command to use `rm -f package-lock.json`

edzkite:
Fixed the `package-lock.json` issue. However, for some reason, `apm install` began to hang. Noticed that the `prepublish` command was deprecated in favor of `prepare` and `prepublishOnly`. (See https://docs.npmjs.com/misc/scripts)

Changing `prepublish` -> `prepublishOnly` was the only way for me to get this to work. @dbratz1177  Let me know if that's ok.

dbratz1177:
@edzkite yeah that should work. thanks

PR #591, opened 2019-04-23
Snippets

edzkite:
Atom Implementation for Snippets.

### Dev Instructions

1. Check out this branch, and run `npm install`.
2. Go to the Atom plugins folder (on Mac it's `~/.atom/packages`) and remove the `kite` folder. [This will need to be done every time the copilot is rebuilt]
3. Return back to the `atom-plugin` folder and run `apm link`. (`npm run clean-dev-install` will ensure a correct env)
4. Open a dev instance of Atom with `atom -d ${path to some folder/file}`.
5. To pick up any code changes (e.g. adding a `console.log`) restart Atom. (`cmd+R` does it)

### DIFF
--- a/lib/data-loader.js
+++ b/lib/data-loader.js
@@ -72,7 +72,7 @@ const DataLoader = {
     const buffer = editor.getBuffer();
     const text = buffer.getText();
     const cursorPosition = buffer.characterIndexForPosition(position);
-    return KiteAPI.getSnippetCompletionsAtPosition(editor.getPath(), text, cursorPosition, 'atom')
+    return KiteAPI.getSnippetCompletionsAtPosition(editor.getPath(), text, 'atom', cursorPosition)
     .then(completions => {
       return completions.map(c => {
         const completion = {
--- a/lib/data-loader.js
+++ b/lib/data-loader.js
@@ -74,36 +74,52 @@ const DataLoader = {
     const cursorPosition = buffer.characterIndexForPosition(position);
     return KiteAPI.getSnippetCompletionsAtPosition(editor.getPath(), text, 'atom', cursorPosition)
     .then(completions => {
-      return completions.map(c => {
-        const completion = {
-          text: c.snippet.text,
-          displayText: c.display,
-          className: 'kite-completion',
-        };
-        if (c.snippet.placeholders.length > 0) {
-          var snippetStr = c.snippet.text;
-          var offset = 0;
-          for (var i = 0; i < c.snippet.placeholders.length; i++) {
-            let placeholder = c.snippet.placeholders[i];
-            placeholder.begin += offset;
-            placeholder.end += offset;
-            snippetStr = snippetStr.slice(0, placeholder.begin)
-            + '${' + (i + 1).toString() + ':' + snippetStr.slice(placeholder.begin, placeholder.end) 
-            + '}' + snippetStr.slice(placeholder.end);
-            offset += 5;
-          }
-          // Strip () from being inserted into buffer
-          completion.snippet = snippetStr.slice(1, snippetStr.length - 1);
+      var parsedCompletions = [];
+      completions.forEach(function(c) {
+        parsedCompletions.push(DataLoader.parseCompletion(editor, c, ''));
+        if (c.children) {
+          c.children.forEach(function(child) {
+            parsedCompletions.push(DataLoader.parseCompletion(editor, child, '...'));
+          });
         }
-        completion.type = c.hint;
-        completion.rightLabelHTML = kindLabel(c.hint);
-        if (c.documentation.text) {
-          completion.descriptionHTML = appendCompletionsFooter(c, escapeHTML(c.documentation.text));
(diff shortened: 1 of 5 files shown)


lib/data-loader.js
```diff
@@ -68,6 +68,62 @@ const DataLoader = {
...
+    const buffer = editor.getBuffer();
+    const text = buffer.getText();
+    const selectionRange = editor.getSelectedBufferRange();
+    const start = buffer.characterIndexForPosition(selectionRange.start);
+    const end = buffer.characterIndexForPosition(selectionRange.end);
+    return KiteAPI.getSnippetCompletionsAtPosition(editor.getPath(), text, 'atom', start, end)
+    .then(completions => {
+      var parsedCompletions = [];
+      completions.forEach(function(c) {
+        parsedCompletions.push(DataLoader.parseCompletion(editor, c, ''));
```
dbratz1177: more idiomatic to do `this.parseCompletions`

dbratz1177: granted, you'd need to do `...forEach((c) => {
    this.parseCompletion(...)
})` 
to get proper scoping...

lib/kite.js
```diff
@@ -432,6 +432,22 @@ const Kite = module.exports = {
         ? list.suggestionListElement
         : atom.views.getView(list);
 
+      atom.packages.activatePackage('snippets').then(snippetsPkg => {
```
dbratz1177: Could you add a comment here about this? Could be useful if we ever need to tweak behavior in an analogous way

lib/kite.js
```diff
@@ -461,6 +477,31 @@ const Kite = module.exports = {
           return safeConfirm.call(manager, suggestion);
         };
 
+        manager.replaceTextWithMatch = (suggestion) => {
```
dbratz1177: Do we not need a `safeReplaceTextWithMatch`?

dbratz1177: You would probably want to use it [here](https://github.com/kiteco/atom-plugin/blob/8f73cf98264e385d8d29e0605fa4223be8ea59d5/lib/kite.js#L465) in the disposal function in order to reset the default behavior on the termination of our plugin

lib/kite.js
```diff
@@ -529,11 +570,12 @@ const Kite = module.exports = {
             if (desc) {
               if (suggestions.length === 0) {
                 desc.style.display = 'none';
+                // TODO: Test when suggestions > 0
```
dbratz1177: what's this for?

edzkite:
> See some of my other comments for nits
> 
> I think that organizationally, `parseCompletion` should be in `lib/kite-data-utils`
> 
> I'm also getting a bug with the selection of keyword arg completions that [this video](https://drive.google.com/open?id=1loHJxWeSyBSY5bmYQAC0D_S8RRgYTuP1) demonstrates
> 
> Beyond those items, though, this looks good to be both from a code and behavior perspective!

Awesome! Couple follow up questions:

1. I see a file called `utils.js` that feeds into `kite-data-utils`. Any particular reason why that's the case?
2. For the video is the bug that you have to tab multiple times when a selection is made, or that the suggestion is repeated? The former is on the radar, and for the latter, you may have to pull the latest backend changes from `FEAT-pythoncomplete-speculate`.

dbratz1177:
1. Not being the architect of this plugin I can only speculate, but it seems that `kite-data-utils` is largely concerned with data normalization from the Kite engine, whereas `utils` is a bit more general purpose / low-level (granted, I can see that the actual constituents `utils` blurs that line a bit)
2. It was that when I was hitting `Enter` on a highlighted completion, it wasn't actually filling in the `...`

lib/kite.js
edzkite: So I don't know if this is just transient, but when I had `safeReplaceTextWithMatch` in the subscription block, it would cause the suggestions experience to lag slightly. Given that it was earlier in the implementation cycle, I'll see if it still occurs.

lib/kite.js
edzkite: Forgot to delete. Removing.

lib/data-loader.js
edzkite: Moving this to `kite-data-utils`

edzkite:
Will have some changes w.r.t adding a config to gate the snippets feature.

Also, @dbratz1177 having trouble repro-ing the issue you found. Did you pull the latest changes from Naman's branch (FEAT-pythoncomplete-speculate) and re-build Kited?

metalogical:
It looks like there's a bug where if no completions are returned from the API, the previous completions are still shown.

You can see this on the latest version of my branch, type `np.array(|)` slowly, and `array_equal` shows up.

Separately I'm digging into why I return no completions in this case.

edzkite:
I'm seeing similar behavior in the old completions experience as well. Note the bottom right, which displays the completions response (`[]`). Open to tweaking the UI as a follow up though.

Screen Shot 2019-05-08 at 10 14 57 AM

PR #596, opened 2019-05-07
Copilot docs engine settings hide pop pat

dbratz1177:
fixes the two listed bugs

### DIFF
--- a/lib/elements/kite-links.js
+++ b/lib/elements/kite-links.js
@@ -2,6 +2,7 @@
 
 const {CompositeDisposable, Emitter} = require('atom');
 const LinkScheme = require('../link-scheme');
+const { copilotProtocolSupported } = require('../utils');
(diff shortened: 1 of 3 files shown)


lib/elements/kite-links.js
```diff
@@ -55,7 +56,9 @@ class KiteLinks extends HTMLElement {
         case 'open-search-docs':
           emitAction && this.emitter.emit('did-click-action-link');
           const path = url.path ? url.path : '';
-          atom.applicationDelegate.openExternal(`http://localhost:46624/clientapi/desktoplogin?d=/python/docs${path}`);
+          copilotProtocolSupported()
+            ? atom.applicationDelegate.openExternal(`kite://docs${path}`)
```
edzkite: Could you add a comment that describes the different behaviors for the two URLs?

PR #608, opened 2019-05-22
Use web_id in Docs link for snippets

edzkite:
Resolves https://github.com/kiteco/kiteco/issues/8087

### DIFF
--- a/lib/kite-data-utils.js
+++ b/lib/kite-data-utils.js
@@ -269,10 +269,15 @@ const symbolId = (symbol) =>
 const appendCompletionsFooter = (completion, descriptionHTML) => {
   const id = completion.symbol && completion.symbol.value &&
       completion.symbol.value[0] && completion.symbol.value[0].id;
-  const actions = id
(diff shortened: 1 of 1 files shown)


lib/kite-data-utils.js
```diff
@@ -269,10 +269,15 @@ const symbolId = (symbol) =>
 const appendCompletionsFooter = (completion, descriptionHTML) => {
   const id = completion.symbol && completion.symbol.value &&
       completion.symbol.value[0] && completion.symbol.value[0].id;
-  const actions = id
+  let actions = id
     ? `<a href="kite-atom-internal://open-search-docs/${idToDottedPath(id)}">Docs<span class="kite-link-icon">&#8663;</span></a>`
     : '';
-
+  // Use web_id if completion is from snippets endpoint
+  if (atom.config.get('kite.enableSnippets')) {
```
dbratz1177: I believe this should be in effect regardless if snippets are enabled

PR #609, opened 2019-05-22
Toggle ACP excludes value to filter out duplicate suggestions

edzkite:
As per https://github.com/atom/autocomplete-plus/wiki/Provider-API#defining-a-provider, setting `excludeLowerPriority` to true will exclude suggestions from lower priority providers.

### DIFF
--- a/lib/completions.js
+++ b/lib/completions.js
@@ -13,7 +13,7 @@ const KiteProvider = {
   ].join(', '),
   inclusionPriority: 5,
   suggestionPriority: 5,
-  excludeLowerPriority: false,
+  excludeLowerPriority: true,
   
   // called to handle attribute completions
   getSuggestions(params) {


adamsmith:
Just to clarify, does this exclude lower priority results only when they're duplicates?

dbratz1177:
My reading of the docs is that it'll filter them out regardless if they're duplicates or not.

edzkite:
Good catch! Unfortunately, this would exclude all results from providers with a lower priority.

I took a stab at:
1. Reverting `excludeLowerPriority` to false and Setting `filterSuggestions` to true. (didn't filter out duplicates)
2. Overriding autocomplete-plus's `filterSuggestions` method directly. (autocomplete-python's suggestions were not being grouped together with Kite's for that method).

Let me know your thoughts on the prioritization of this. Otherwise, I can go ahead and revert/merge the PR and then create an issue to track this.

its-dhung:
IMO this is not great behavior if this filters out Atom's native keyword suggestions because that provides value if e.g. Kite's type inference doesn't work on an object. Since we know there are certain libraries / code patterns that cause Kite's type inference to fail, I don't think we should do this. If we're worried about users thinking that Kite shows duplicate suggestions, we should improve our branding so that it's very clear which completions are coming from Kite and which aren't.

adamsmith:
I agree we shouldn’t exclude lower priority completions providers. 

I think overriding filterSuggestions to remove duplicates of Kite completions is something we should do, but it shouldn’t be considered critical or a blocker for releasing anything.

edzkite:
Closing this PR without merging, since this is not the direction we want to go in. Here's the [Issue](https://github.com/kiteco/kiteco/issues/8160) I created to track follow ups for this.

PR #614, opened 2019-05-30
Ksg experiment

dbratz1177:
Implements the initial Keyword Snippets feature for Atom

**Product Notes**:
- Accessible via the command palette and via keyboard shortcut (`cmd/ctrl+alt+l` - though definitely open to changing that)
- Widget aligns with and inserts code at the cursor (inserted text is initially selected)
- Both search and codeblock result dropdowns are navigable with arrow keys
- queries to both the codeblock and search endpoints are prepended (in a non-user facing way) with `python` if `python` is not already in the query

**Technical Notes**
- Basic architecture is essentially a flavor of MVC. There are model and view layers for both Search and Codeblocks. There is a top-level module controller, as well as a separate view controller.
- This feature heavily leverages Event emissions to communicate between layers.
- The view layers heavily leverage the creation of custom `HTMLElement`s
- Testing still needs to be designed and implemented - but that seemed outside the scope of this initial experiment


keymaps/win32.cson
```diff
@@ -3,6 +3,9 @@
   'ctrl-alt-shift-k': 'kite:docs-at-cursor'
   'ctrl-alt-k': 'kite:open-in-web'
 
+'.platform-win32 .workspace':
+  'cmd-alt-l': 'kite:ksg'
```
its-dhung: This should be `ctrl`

keymaps/linux.cson
```diff
@@ -2,6 +2,10 @@
 '.platform-linux atom-text-editor:not([mini])':
   'ctrl-alt-shift-k': 'kite:docs-at-cursor'
   'ctrl-alt-k': 'kite:open-in-web'
+  'ctrl-alt-l': 'kite:ksg'
+
+'.platform-linux .workspace':
+  'cmd-alt-l': 'kite:ksg'
```
its-dhung: This should be `ctrl`

PR #625, opened 2019-08-17
Update Kite statuses

edzkite:
Relevant changes:

- https://github.com/kiteco/atom-plugin/compare/status-update?expand=1#diff-0236e9ad89cb02cd990eabd968e1291dR329
- https://github.com/kiteco/atom-plugin/compare/status-update?expand=1#diff-9971595d5358d9bfdac4915f7b22664bR131

### DIFF
--- a/lib/elements/kite-status-panel.js
+++ b/lib/elements/kite-status-panel.js
@@ -2,15 +2,15 @@
 
 require('./kite-logo');
 
-const {CompositeDisposable, TextEditor} = require('atom');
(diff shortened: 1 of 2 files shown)


lib/elements/kite-status-panel.js
```diff
@@ -299,6 +326,10 @@ class KiteStatusPanel extends HTMLElement {
                 content = `<div class="ready">Kite engine is indexing your code ${dot}</div>`;
                 break;
 
+              case 'noindex':
```
metalogical: `"noIndex"` should be what we're returning

PR #650, opened 2020-02-28
Trigger In-Editor Install Flow for WarnNotInstalled Notification and Status Panel

tonycheang:
Original issue: https://github.com/kiteco/kiteco/issues/9172
Sibling PR: https://github.com/kiteco/kite-installer/pull/91

Previously, the "Install now" button on the status panel linked externally to Kite's download page, and the uninstalled warning was disabled because there were errors related to resolving relative file paths after Webpack finished bundling into dist/main.js.

This PR fixes the associated bugs and has both the notification and the status bar, after clicking "install" or "install now", install the desktop app.

Here's the behavior after the fix when the Kite atom plugin is installed but desktop app is not:

notif-install
status-install

Behavior and some assumptions made:
1) The warn-not-installed notification only comes up if the user has not installed Kite then uninstalled (e.g. first time installation). This was default behavior based off existing code. Do we want this notification to always show up regardless if they've previously uninstalled?
2) Since this now shows a notification that prompts the user to click install, I removed automatic installation of the desktop application when the user downloads the Kite atom plugin.
3) The warn-not-installed notification pops up regardless of whether the user is on an editor with a supported language. Previous behavior required the user to be on an editor with a supported language.
4) The install button on the notification triggered an older 'defaultFlow' that asked a user for their email to continue, but that seems handled in the Desktop application now. I removed it to streamline the download.

@its-dhung What are your thoughts on this (e.g. should install desktop app by default the first time instead of showing a notification)?


config/webpack.config.js
```diff
@@ -17,6 +17,9 @@ const config = {
     libraryTarget: 'commonjs2',
     library: '',
   },
+  resolve: {
+    modules: ['../node_modules']
```
metalogical: Just doing `['node_modules']` here should work. Webpack will automatically walk up the directory tree looking for relative paths.

lib/elements/kite-status-panel.js
```diff
@@ -138,25 +138,20 @@ class KiteStatusPanel extends HTMLElement {
         atom.applicationDelegate.openExternal(`kite://settings/permissions${url.path || ''}`);
         break;
       case 'install':
-        this.pendingStatus('Installing Kite');
+        this.pendingStatus('Working on adding Kite...');
```
metalogical: why change this? I think "Installing Kite" is sufficiently clear.

lib/elements/kite-status-panel.js
```diff
@@ -138,25 +138,20 @@ class KiteStatusPanel extends HTMLElement {
...
-              throw err;
-            }
-          })
+        this.app
+          .installFlow()
           .then(() => {
             this.installing = false;
           })
-          .catch(() => {
+          .catch((err) => {
```
metalogical: if you don't use the parameter, you should simply not have it here.

tonycheang: Remnants from debugging--I'll remove this, thanks!

config/webpack.config.js
tonycheang: Oh that's neat. I'll fix it up.

lib/elements/kite-status-panel.js
tonycheang: Yeah, sufficiently clear is a good point. I was a bit hung up about the download-page shown going from a downloading progress-bar to the same install icon and text that's been spinning in the status panel throughout the process. But it is a minor point. I'll revert this.

its-dhung:
> The warn-not-installed notification only comes up if the user has not installed Kite then uninstalled (e.g. first time installation). This was default behavior based off existing code. Do we want this notification to always show up regardless if they've previously uninstalled?

I think this notification should _always_ show up, mainly because the Kite plugin is useless unless the Kite Engine is also installed. We already have some notifications that always show if the user's environment is configured improperly e.g. the `tree-sitter` warning. This install notification should behave similarly.

> Since this now shows a notification that prompts the user to click install, I removed automatic installation of the desktop application when the user downloads the Kite atom plugin.

👍 

> The warn-not-installed notification pops up regardless of whether the user is on an editor with a supported language. Previous behavior required the user to be on an editor with a supported language.

👍 

> The install button on the notification triggered an older 'defaultFlow' that asked a user for their email to continue, but that seems handled in the Desktop application now. I removed it to streamline the download.

👍

PR #652, opened 2020-03-17
Un-silence errors for acp package.

tonycheang:
Saw while looking through another issue and wondered if there's a lot of output here or if there's a good reason to silence.


edzkite:
One potential reason we suppress errors here is because the output is noisy and does not correspond to an actual error.

Unless it'd add value to a user outside of Kite, the log should be kept locally imo.

PR #653, opened 2020-03-19
Add write-through closing brackets for selected Kite completions.

tonycheang:
Fixes https://github.com/kiteco/kiteco/issues/10318

### What was done

The `BracketTracker` class reads selected kite completions to check for closing brackets and hooks into `editor.onWillInsertText` to check whether a user would like to insert a closing bracket at a location that should write-through.

atom-write-through-parens

This behavior of navigating away, navigating back in, then still writing through is consistent with the default (shipped with the Atom IDE) `bracket-matcher` package behavior.

### QA

Check behavior is like that of the GIF for all supported languages.

### DIFF
--- a/lib/bracket-marker.js
+++ b/lib/bracket-marker.js
@@ -1,0 +1,79 @@
+const {CompositeDisposable} = require('atom')
+const _ = require('lodash')
+
+const BracketPairs = ['()', '[]', '{}', '""', "''", '``', '“”', '‘’', '«»', '‹›']
+const ClosingBrackets = new Set(BracketPairs.map(x => x[1]))
+
+module.exports =
+class BracketMarker {
+  init() {
+    this.bracketMarkersByEditorID = {}
+    const observe = atom.workspace.observeTextEditors(editor => {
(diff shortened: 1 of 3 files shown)


lib/bracket-marker.js
```diff
@@ -0,0 +1,79 @@
...
+    /* Marks the brackets as 'write-through-able' after an insertion. */
+
+    const bracketMarkers = this.bracketMarkersByEditorID[editor.id]
+    // Uninitialized edtior won't check to skip closing brackets; pointless to mark.
+    if (!bracketMarkers) return
+
+    for (let i = 1; i < text.length; i++) {
+      if (this.isClosingBracket(text.charAt(i))) {
+        const range = [cursorPosBeforeInsert.traverse([0, i-1]), cursorPosBeforeInsert.traverse([0, i])]
+        const marker = editor.markBufferRange(range)
```
metalogical: We may want to register a callback with `marker.onDidDestroy()` so that we can remove dead markers from our `bracketMarkers` list, even if the marker was destroyed by the editor. Then we shouldn't remove from `bracketMarkers` below.

PR #654, opened 2020-04-04
center Kite icon for lexical completions

metalogical:
closes kiteco/kiteco#10614

lmk if there's a better way to center the icon.

image

I take advantage of the fact that lexical returns an empty hint, while semantic Python always returns "unknown." If this is approved, I will document this contract in the completion API docs as well as the editor plugin docs.


styles/kite-completions.less
```diff
@@ -19,6 +19,21 @@ autocomplete-suggestion-list {
...
 
+              &.kite-empty-type {
+                background: fadeout(@text-color, 90%); // TODO(naman) inherit from AC+ styles
+                .icon-letter {
+                  visibility: hidden;
+                }
+                &::after {
+                  mask-image: url('atom://kite/assets/logo-small.svg');
+                  -webkit-mask-image: url('atom://kite/assets/logo-small.svg');
+                  bottom: 50%
```
tonycheang, line 30: nit question: does this work without a terminating semi-colon?

metalogical: good catch! fixing this makes the numbers work out much more nicely

PR #655, opened 2020-04-04
send "trigger" type to backend

metalogical:
a string: one of `manual`, `automatic`, `snippet`.

If this is approved, I will document this in the API docs so that other editors follow the same scheme.


lib/kite.js
```diff
@@ -572,7 +572,8 @@ const Kite = (module.exports = {
             if (expansion && expansion.goToNextTabStop()) {
               nextTabStopVisited = true;
             }
-            manager.requestNewSuggestions();
+            manager.shouldDisplaySuggestions = true;
```
edzkite, line 575: `requestNewSuggestions` takes into account if the user has configured a [delay](https://github.com/atom/autocomplete-plus/blob/master/lib/autocomplete-manager.js#L688), defined as [follows](https://github.com/atom/autocomplete-plus/blob/master/lib/autocomplete-manager.js#L688).

We may want to respect that behavior here, even if the default value is 0.

PR #659, opened 2020-04-15
Patch autocomplete-plus/snippets insert(...) to not indent Kite completions.

tonycheang:
Fixes https://github.com/kiteco/kiteco/issues/10726
Fixes https://github.com/kiteco/kiteco/issues/10656

### What was done
Removed indentation insertion for multiline completions. The crux of the change revolves around the constructor of `lib/snippets/snippet-expansion.js`. A flag `shouldNotIndent` was added to the arguments. (compare with [original](https://github.com/atom/snippets/blob/36054ba644a4b1f7ea2ad62f102543f97fac4e15/lib/snippet-expansion.js#L4))

#### After
atom_mulit_after
atom_multiline_py

See issues linked for behavior before.

### Considerations
We probably really only need to patch the constructor of `SnippetExpansion` and somehow ensure autocomplete-plus is using an `insert` that returns that instead of copying `SnippetExpansion` and all of `Snippet`'s dependencies but `package.mainModule` doesn't seem to include class constructors. 

Potentially the changes to add a `shouldNotIndent` flag can be proposed as changes to atom/snippets and autocomplete-plus to avoid patching.

### QA
Ensure completions work almost exactly as before except that during the insertion of a completion containing a newline, the indentation level matches what kited sends as a snippet (no regression and no additional indentation added).


tonycheang:
> It would be slightly cleaner to put all the custom snippets code into a fork of atom/snippets: kiteco/snippets.
> 
> And then publish our `kite-atom-snippets` npm (apm?) package, and then depend on that package.
> 
> That would allow us to track exactly how our version differs from upstream, and also contribute to upstream later if we want.

Yeah, I was debating between that. I saw that their package wasn't on NPM and wasn't sure the etiquette around that. Should I go ahead and make an npm package of a fork instead?

metalogical:
I'm not sure if `apm` is the way to go actually. That is used mostly for independent Atom packages. If we released our forked snippets package as a separate Atom packages, we would need to be able to:

- specify that Kite depends on KiteSnippets
- installing KiteSnippets doesn't conflict with the default Snippets package

It may be better to just publish on `npm` and use an `npm` dependency, since we don't really intend for anyone to install our snippets fork separately.

cc @edzkite

metalogical:
The other option is to just add our snippets fork as a Git submodule for now, and then we don't need to release that anywhere.

edzkite:
+1 for adding the fork as a submodule.

lib/kite.js
```diff
@@ -669,7 +687,8 @@ const Kite = (module.exports = {
               const cursorPosBeforeInsert = editor.getCursorBufferPosition();
               // Insert suggestion
               if (suggestion.snippet && manager.snippetsManager) {
-                manager.snippetsManager.insertSnippet(suggestion.snippet, editor, cursor);
+                const shouldNotIndent = suggestion && suggestion.className && suggestion.className == 'kite-completion';
```
edzkite, line 690: suggestion.className ===?

PR #661, opened 2020-04-24
Atom Onboarding with Command Palette

tonycheang:
Addresses Atom for kiteco/kiteco#10456 for the updated spec.

### What was done
Essentially leveraged the second argument in `KiteAPI.getOnboardingFilePath` to pull language specific tutorial files. They don't yet exist so this change would at the moment create a warning notification. ~Should test and merge after those files get added to the endpoint.~ Should merge for release at JS public launch.

### QA
Check that:

1. Opening Atom with `kite.showWelcomeNotificationOnStartup: true` in atom's `config.cson` and `"has_done_onboarding": "false"` in `~/.kite/settings.json` attempts to open the Python onboarding file. (no regression)
2. Commands for `Kite: Go Tutorial` and `Kite: Javascript Tutorial` are present in the command palette. (Note that these will fail to pull an onboarding file from kited at the moment as those files do not exist).

### DIFF
--- a/lib/kite.js
+++ b/lib/kite.js
@@ -271,7 +271,9 @@ const Kite = (module.exports = {
     this.subscriptions.add(
       atom.commands.add('atom-workspace', {
         'kite:search-stack-overflow': () => this.toggleKSG(),
-        'kite:tutorial': () => this.openKiteTutorial(true),
+        'kite:python-tutorial': () => this.openKiteTutorial(true, 'python'),
+        'kite:javascript-tutorial': () => this.openKiteTutorial(true, 'javascript'),
+        'kite:go-tutorial': () => this.openKiteTutorial(true, 'go'),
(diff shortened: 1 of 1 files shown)


plungg:
@tonycheang I tested this PR, commands opened tutorials correctly, but the tutorial did not open on start.

This could likely be due to a setup issue on my end. You can see my test in [this linked video](https://www.dropbox.com/s/k1py3zm52iclv14/atom-onboarding.mov?dl=0) where my `npm run build-dev` step had errors.

tonycheang:
@plungg I can reproduce if I change ~/.kite/settings.json and open atom without restarting Kite. I believe we only read from settings.json once on startup. Did you change the value while Kite was open / have you tried restarting Kite first?

plungg:
@tonycheang that fixed it!

PR #687, opened 2020-09-09
Set selection object encoding.

### LINKED ISSUE #686: Edit events send incorrect encoding for selections
The selection objects sent in an edit event have `selection.Encoding` set to `utf-8`, which causes this error: `process event error: could not decode selection start offset: offset not aligned to code point boundary`. The events are utf-16 encoded, so this should be the value set.

hrysoula:
Fixes #686 

This PR sets the encoding on the selection object sent with edit events to `utf-16`.


hrysoula:
@metalogical I can't find instructions for testing the plugin anywhere. I've attempted updating `~/.atom/packages/kite/` with the changes, but this doesn't seem to be working. Any advice?

metalogical (approved):

PR #688, opened 2020-09-25
Remove hard coded max file size checks

edzkite:
Unblocks kiteco/kiteco#11539

Update Kite Status panel and tooltip to check Kited value, remove check from event and hardcoded values.

Would recommend toggling Hide Whitespace changes due to eslint autoformatting.

Dependent on https://github.com/kiteco/kite-api-js/pull/43


lib/data-loader.js
```diff
@@ -140,12 +138,16 @@ const DataLoader = {
     return KiteAPI.getUserAccountInfo();
   },
 
+  getMaxFileSize() {
```
tonycheang, line 141: nit: `getMaxFileSizeBytes`?

tonycheang (approved):

adamsmith:
I couldn't tell from briefly reading the diff, but we should make sure we don't JSON serialize file buffers when the length is over the max file size. Doing that used to cause huge performance issues in our Atom plugin since the action is triggered on each keystroke. (This comment applies to other plugins also.)

metalogical:
> I couldn't tell from briefly reading the diff, but we should make sure we don't JSON serialize file buffers when the length is over the max file size. Doing that used to cause huge performance issues in our Atom plugin since the action is triggered on each keystroke. (This comment applies to other plugins also.)

This is a good point. @edzkite if we're going to do the file size checks only on the backend, we should ensure the buffer contents is sent as raw bytes (requiring no JSON-serialization to a string) on the socket. This might involve changing how the event / completions request is sent to the backend, so that may end up being excessively complicated.

PR #691, opened 2020-10-02
status initializing

tonycheang:
Addresses https://github.com/kiteco/kiteco/issues/11798 for atom

### What was done
Mostly refactored the tool tip and status icons to directly use what kited returns instead of switching on the status.status. It seems like the messages diverge between plugins because it's hard to remember to sync the tooltip and innerhtml when the api changes, though any new statuses must still be added to the stylesheet to put the correct icon in the bar. Maybe we can change the api a bit to avoid this, but I'm not sure there's a clear spec on this since it seems editor specific?

The effect of this change would change most text from `'Kite engine is...'` to `'Kite is...'`

Same question as the vscode issue: is 5 seconds frequent enough for a 15 second warmup?


metalogical (approved):
I think the refresh interval of 5s is fine for now. I think it makes most sense to have a socket open to Kite and have Kite push the update (same for copilot UI state). We may end up implementing this for predictive navigation (or not). But for now, this is good enough.

PR #692, opened 2020-10-06
add all-language-lexical languages

ryantberg:
kiteco/kiteco#11791

### DIFF
--- a/lib/editors.js
+++ b/lib/editors.js
@@ -6,18 +6,84 @@ const KiteEditor = require('./kite-editor');
 const KiteAPI = require('kite-api');
 
 const EXTENSIONS_BY_LANGUAGES = {
-  python: [
-    'py',
+  bash: [
+    'sh',
+  ],
+  c: [
+    'c',
+    'h',
+  ],
+  cpp: [
+    'cc',
+    'cpp',
+    'hpp',
+  ],
+  csharp: [
+    'cs',
+  ],
+  css: [
+    'css',
   ],
   go: [
     'go',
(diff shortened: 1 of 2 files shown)


lib/editors.js
```diff
@@ -6,19 +6,85 @@ const KiteEditor = require('./kite-editor');
 const KiteAPI = require('kite-api');
 
 const EXTENSIONS_BY_LANGUAGES = {
```
edzkite, line 8: Is this list standard across all the editors? Any reason to not expose an endpoint in Kited to get this list?

edzkite (changes requested):

lib/editors.js
ryantberg: I believe the issue with that is if the Kite Engine isn't installed/running, the status bar wouldn't be present to inform the user of that.

edzkite: Good point.

However, if we're switching to an all language world, wouldn't it make more sense to invert the logic so that the Kite status is always visible, unless the current file type is unsupported?

With the current approach, updating the languages we support means we would have to go through all our editors and update them individually. Ideally, we should minimize the chance of small diffs in behavior between editors.

On the other hand, we could do more work upfront, but save a lot of headaches in the future.

ryantberg: I think that approach requires enough investigation that it should be considered separately from the ALL release, and I'm not sure if it is worth the complication unless we start changing language support dynamically or at a much higher frequency than we currently do.

PR #693, opened 2020-10-06
Remove go beta notif

tonycheang:
Addresses https://github.com/kiteco/kiteco/issues/11884

I don't think this was actually being called anywhere, but since I was there, might as well remove it.


lib/notifications-center.js
```diff
@@ -146,39 +146,7 @@ class NotificationsCenter {
     this[this.NOTIFIERS[state]] && this[this.NOTIFIERS[state]](state);
   }
 
-  showGoBetaNotification(kite) {
-    if (this.hasSeenGoBetaNotification || !atom.config.get('kite.showGoBetaNotification')) {
```
edzkite, line 150: We should remove `hasSeenGoBetaNotification` from the constructor as well.

PR #694, opened 2020-10-07
Remove completion request on placeholder selection

tonycheang:
Addresses https://github.com/kiteco/kiteco/issues/11202 for Atom

This should apply to all snippets from all providers because we patch autocomplete-plus's snippets manager to insert our own behavior (cc @edzkite is this true based off your understanding?).

The regex is a bit nasty. We could potentially return a flag from kited instead, but thought I'd get this out for discussion to see if we want to address on-select completions in-editor vs kited for Atom first. Unsure if we always return `1` for the first one/whether to rely on that if we think regex is ok for this.

behavior

This will request completions when selecting `len(filter${1:})$0`. Not sure if that's what we want either.

### DIFF
--- a/lib/kite.js
+++ b/lib/kite.js
@@ -545,14 +545,10 @@ const Kite = (module.exports = {
       if (atom.config.get('kite.enableSnippets')) {
         atom.packages.activatePackage('snippets').then(snippetsPkg => {
           const snippets = snippetsPkg.mainModule;
-          const safeGoToNextTabStop = snippets && snippets.goToNextTabStop;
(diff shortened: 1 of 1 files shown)


metalogical:
I'm not sure we actually care to distinguish between empty vs non-empty placeholder selections. The fix for IntelliJ does not make this distinction. It may be worth bringing that up in the underlying issue, but in the absence of a strong product reason for doing so, I would vote to simplify this and just consider all placeholder selections the same.

edzkite (approved):

PR #696, opened 2020-10-12
Fix TypeError: i.then(...).then(...).catch(...).finally is not a function

edzkite:
May resolve https://github.com/kiteco/issue-tracker/issues/418

The [previous implementation](https://github.com/kiteco/atom-plugin/blob/155ee06cbf1dc0109a92f76c0b95297d62e4725e/lib/editor-events.js#L93) returned a `Promise<void>`.

The [current implementation](https://github.com/kiteco/atom-plugin/blob/b04a0520186785c28a4ee3356ca9c82ac048704b/lib/editor-events.js#L88) does not return anything.


ryantberg:
(reposting from Slack)

I think it's actually replacing the last `then` with `finally`. That [isn't supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally#Browser_compatibility) until Node 10, and they're running Atom `1.26.`1 which as far as I can find so far seems to be running with an earlier version.

tonycheang (approved):

PR #698, opened 2020-10-13
Rotate Rollbar token

edzkite:


metalogical (approved):
once this is released, we can disable the old access token

PR #700, opened 2020-10-23
Find Related Code (experimental)

tonycheang:
Addresses https://github.com/kiteco/kiteco/issues/11900 for atom

Mostly the same as the vscode implementation, but I had some trouble passing functions to `then` and `catch` directly, e.g. `prom.then(this.somefunc)` fails but `prom.then((arg) => this.somefunc(arg))` works. Is that related to how we have our bundling currently set up?


edzkite (approved):

edzkite:
> Mostly the same as the vscode implementation, but I had some trouble passing functions to `then` and `catch` directly, e.g. `prom.then(this.somefunc)` fails but `prom.then((arg) => this.somefunc(arg))` works. Is that related to how we have our bundling currently set up?

If `somefunc` uses an argument, how would it get passed in `prom.then(this.somefunc)`?

tonycheang:
afaik, `((arg) => this.someFunc(arg))(someArg)` is no different than calling `(this.someFunc)(someArg)`, where
```js
class SomeClass {
  someFunc(arg) {
    ...
  }
}
```
until I ran across the discrepancy here. I'd expect `.then()` to call the passed-in function with whatever the previous promise resolved with. This seems to be the case for our vscode plugin.

tonycheang:
Maybe a candidate for our JS pitfalls discussion.

https://stackoverflow.com/a/34930859
[YDKJS](https://github.com/getify/You-Dont-Know-JS/blob/1st-ed/es6%20%26%20beyond/ch2.md#not-just-shorter-syntax-but-this): "[Arrow functions] lexically inherits this from the surrounding scope."

PR #701, opened 2020-10-25
Improve the performance of get suggestions list element 

aminya:
This improves the performance of `get suggestions list element` by caching the values related to autocomplete-plus.

### Benchmarks

Previously Kite was calling quite expensive API such as `atom.packages.getAvailablePackageNames` on every single keystroke. This has caused the performance of the text editor to be reduced.
image

This PR fixes this issue by caching the result. This 620ms is completely gone now
image

### DIFF
--- a/lib/completions.js
+++ b/lib/completions.js
@@ -5,6 +5,7 @@ const KiteSignature = require('./elements/kite-signature');
 const { delayPromise } = require('./utils');
 let Kite;
 
+let isAutocompletePlusInstalled = null;
 const KiteProvider = {
   selector: '.source, .text',
   disableForSelector: [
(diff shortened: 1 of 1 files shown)


lib/completions.js
```diff
@@ -158,19 +161,27 @@ const KiteProvider = {
...
+    }
+    // return if not installed
+    if (isAutocompletePlusInstalled === false) {
       return null;
     }
     if (this.suggestionListElement) { return this.suggestionListElement; }
 
-    const pkg = atom.packages.getActivePackage('autocomplete-plus').mainModule;
-    if (!pkg || !pkg.autocompleteManager || !pkg.autocompleteManager.suggestionList) {
+    // get autocomplete-plus pkg
```
tonycheang, line 174: nit: let's remove the comments (check...installation, return..., get...) since it's fairly apparent what's going on via naming and code structure

tonycheang (commented):
Thanks for the contribution! Just one minor nit and we can merge this.

lib/completions.js
aminya: Fixed in https://github.com/kiteco/atom-plugin/pull/701/commits/883af3f9b1761951e8af47e8d1ddaababd226a73

PR #708, opened 2020-12-09
Handle ErrPathHasUnsupportedExtension

edzkite:
Screen Shot 2020-12-09 at 12 07 40 PM

Atom implementation of kiteco/kiteco#12098.

### DIFF
--- a/lib/notifications-center.js
+++ b/lib/notifications-center.js
@@ -510,9 +510,7 @@ class NotificationsCenter {
         switch (responseData.trim()) {
           case 'ErrPathHasUnsupportedExtension':
             this.queue.addWarning(KiteCodeFinderError, {
-              description: [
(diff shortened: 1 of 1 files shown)


lib/notifications-center.js
```diff
@@ -507,6 +508,19 @@ class NotificationsCenter {
 
       if (responseData && typeof responseData === 'string') {
         switch (responseData.trim()) {
+          case 'ErrPathHasUnsupportedExtension':
+            this.queue.addWarning(KiteCodeFinderError, {
+              description: [
+                `Code Finder does not support the \`${path.extname(filename)}\` file extension yet.`,
+              ].join(' '),
```
tonycheang, line 515: nit: I think if we go with template literal, we don't need the `[].join(' ')`

tonycheang (approved):

PR #710, opened 2020-12-18
Related Code Error Handling Improvements

edzkite:
Atom portion of https://github.com/kiteco/kiteco/issues/12248

In addition to removing empty line checks:
- Cleaned up `getRelatedCodeErrHandler`
- Fix error handling when there's no active text editor

### DIFF
--- a/lib/codenav.js
+++ b/lib/codenav.js
@@ -18,17 +18,15 @@ module.exports = class Codenav {
   }
 
   relatedCodeFromFileWithEditor(textEditor) {
-    KiteAPI
-      .requestRelatedCode('atom', atom.packages.getApmPath(), textEditor.getPath(), null, null)
-      .catch(this.notifications.getRelatedCodeErrHandler());
+    return KiteAPI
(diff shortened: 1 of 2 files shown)


lib/codenav.js
```diff
@@ -7,38 +7,28 @@ module.exports = class Codenav {
...
-      .then((textEditor) => this.relatedCodeFromFileWithEditor(textEditor));
+      .then((textEditor) => this.relatedCodeFromFileWithEditor(textEditor))
+      .catch(this.notifications.getRelatedCodeErrHandler());
   }
 
   relatedCodeFromLine() {
     this.requireEditor()
-      .then((textEditor) => this.relatedCodeFromLineWithEditor(textEditor));
+      .then((textEditor) => this.relatedCodeFromLineWithEditor(textEditor))
+      .catch(this.notifications.getRelatedCodeErrHandler());
```
tonycheang, line 17: Oh I suppose the catch out here makes sense since `requireEditor()` can throw, good catch. I think means in the `...WithEditor` methods we can `return KiteAPI.requestRelatedCode(...)` to propagate the request errors, right?

tonycheang (approved):

PR #711, opened 2021-01-07
Codenav line decoration

tonycheang:
Addresses https://github.com/kiteco/kiteco/issues/11934 for Atom

### What was done
- Added a click-able line decoration for predictive navigation when it's enabled
Screen Shot 2021-01-06 at 4 44 21 PM

@its-dhung Lmk how this looks, changes you'd make, or if you want to take a pass on the styling

### DIFF
--- a/lib/codenav-decoration.js
+++ b/lib/codenav-decoration.js
@@ -1,0 +1,95 @@
+const KiteAPI = require('kite-api');
+const KiteCodenavButton = require('./elements/kite-codenav-button');
+
+module.exports = class CodenavLineDecoration {
+  constructor(onClick) {
+    this.onClick = onClick;
+    this.activeEditor = null;
+    this.marker = null;
+    this.decoration = null;
+    this.lineInfo = null;
+
+    atom.workspace.observeTextEditors(editor => {
+      editor.onDidChangeCursorPosition(this.onDidChangeCursorPosition.bind(this));
+    });
+    atom.workspace.onDidChangeActiveTextEditor(this.onDidChangeCursorPosition.bind(this));
+    this.onDidChangeCursorPosition();
+  }
+
+  dispose() {
+    this.marker && this.marker.destroy();
+  }
+
+  enabled() {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -24,6 +24,12 @@
       "title": "Show welcome notification on startup",
       "description": "Whether or not to show the Kite welcome notification on startup."
     },
+    "enableCodeFinderLineDecoration": {
+      "type": "boolean",
+      "default": "true",
```
edzkite, line 29: Should we set the default here to false until Codenav is released? 

Alternatively, we should ensure this change doesn't get released in the Atom plugin until the feature flags are removed.

lib/elements/kite-codenav-button.js
```diff
@@ -0,0 +1,41 @@
+function KiteCodenavButton(message, onClick, {
+  charWidth,
+  lineHeightPx,
+  lineLengthPx,
+}) {
+  // <div class="kite-codenav-decoration">
```
edzkite, line 6: Remove?

lib/codenav-decoration.js
```diff
@@ -0,0 +1,96 @@
...
+          }
+        ),
+      });
+    }
+  }
+
+  async reset(editor) {
+    this.dispose();
+    this.activeEditor = editor;
+    this.lineInfo = undefined;
```
edzkite, line 73: Nit: `lineInfo` is set to `null` in the constructor.

package.json
tonycheang: As is, the endpoint will 501, leading to no decoration, but requests will be continually made. This is a little wasteful, but it makes QA a bit easier. I'll remove error logging to not spam the console.

lib/elements/kite-codenav-button.js
tonycheang: The structure seems a bit clearer to me with a comment rather than having to reconstruct it while reading the code below

PR #715, opened 2021-02-02
All Feat Pro Notification

tonycheang:
Addresses https://github.com/kiteco/kiteco/issues/12288 for Atom

### What was done
- Added notification error handler that reads and shows a notification if it exists
- Refactored the docs command and related logic to live with the hover provider in `lib/docs.js`

### DIFF
--- a/lib/docs.js
+++ b/lib/docs.js
@@ -2,8 +2,8 @@ const DataLoader = require('./data-loader');
 const VirtualCursor = require('./virtual-cursor');
 
 export class DocsCommand {
-  constructor(notifyfromError) {
-    this.notifyfromError = notifyfromError;
+  constructor(notifCenter) {
+    this.notifCenter = notifCenter;
   }
 
   dispose() {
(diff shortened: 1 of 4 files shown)


lib/docs.js
```diff
@@ -0,0 +1,69 @@
+const DataLoader = require('./data-loader');
+const VirtualCursor = require('./virtual-cursor');
+
+export class DocsCommand {
+  constructor(notifyfromError) {
+    this.notifyfromError = notifyfromError;
```
edzkite, line 6: Is there any reason why we pass in just `notifyFromError` here as opposed to the [notifications center directly](https://github.com/kiteco/atom-plugin/pull/715/files#diff-ad169180a7e1b63584cc63f70b896728043da578d51d0b80fe653cb0de89b50cR8)?

tonycheang: Well, not a strong opinion, but it makes explicit the only dependency is this particular function. The counter argument would be if we transition to typescript one day, passing the notifications center directly means we can just add a type annotation. Do you have a preference?

edzkite: Makes sense. Imo, simplicity would be preferable here, so I'd pass it in directly.

edzkite (approved):

PR #718, opened 2021-02-05
Status Panel Button And Styling

tonycheang:
Addresses https://github.com/kiteco/kiteco/issues/12288 's status panel for Atom

## What was done
- Main use case is for the `all_features_pro` button and text but since I needed specific alignment,
- Refactored various statuses to make things more uniform

### Commentary
The login button is suppressed when the status sends over a button. Periods indicate splitting (`Locked until tomorrow. Upgrade now to unlock.`). The spec had the button on the left, though most of Atom's buttons are on the right. I think this looks nice though.
Screen Shot 2021-02-04 at 3 11 41 PM

If a button isn't sent but a message with periods is sent, we'd get something like this, though currently we only send a message with a period with this particular status and we always send a button with it as well. Separately, IMO the login button looks misplaced here and other mockups, but not too sure how to fix.
Screen Shot 2021-02-04 at 3 33 04 PM

I saw that the spec had this as red, so I kept it as red, but it seems to me it would look better yellow, since it's a 'pending' kind of message.
Screen Shot 2021-02-04 at 2 48 49 PM

### More
Screen Shot 2021-02-04 at 3 08 24 PM
Screen Shot 2021-02-04 at 3 08 14 PM
Screen Shot 2021-02-04 at 3 05 18 PM
Screen Shot 2021-02-04 at 2 58 06 PM

### Deprecated(?)
I don't really have the context around these, but our terminology seems different now. Should we remove these? There's a lot of related code lying around, and I didn't want to remove everything at once.
Screen Shot 2021-02-04 at 2 56 19 PM
Screen Shot 2021-02-04 at 2 53 56 PM

### DIFF
--- a/lib/elements/kite-status-panel.js
+++ b/lib/elements/kite-status-panel.js
@@ -260,10 +260,8 @@ class KiteStatusPanel extends HTMLElement {
     switch (status.state) {
       case STATES.UNSUPPORTED:
         content = `
-          <div class="center-aligned-line">
-            <div class="text-danger">
-              Kite engine is not available on your system
-            </div>
+          <div class="center-aligned-line text-danger">
+            Kite engine is not available on your system
             ${dot('text-danger')}
           </div>
         `;
@@ -367,8 +365,8 @@ class KiteStatusPanel extends HTMLElement {
             break;
           case editor && editor.getText().length >= maxFileSize:
             content = `
-              <div class="center-aligned-line">
-                <div class="text-warning">The current file is too large for Kite to handle</div>
+              <div class="center-aligned-line text-warning">
+                The current file is too large for Kite to handle
                 ${dot('text-warning')}
               </div>
             `;
--- a/lib/elements/kite-status-panel.js
+++ b/lib/elements/kite-status-panel.js
@@ -10,7 +10,6 @@ const LinkScheme = require('../link-scheme');
 const DataLoader = require('../data-loader');
 
 const { STATES } = KiteAPI;
-const dot = '<span class="dot">•</span>';
 
 class KiteStatusPanel extends HTMLElement {
   constructor(kite) {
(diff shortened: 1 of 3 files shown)


lib/elements/kite-status-panel.js
```diff
@@ -256,64 +255,154 @@ class KiteStatusPanel extends HTMLElement {
...
             }
-          }
+            // The long status is split at periods. Each line will get their own row
+            // So it's best not to have more than two sentences in the long status
+            const textLines = syncStatus.long
+              .split('.')
+              .filter(content => content !== '')
+              .map(content => `<div class="text-line">${content}</div>`)
+              .join('');
+            const color = syncStatus.long.includes('ready') ? 'ready' : 'text-warning';
```
tonycheang, line 391: This is kind of a hack to avoid sending color information from kited as well, since I haven't thought about how that should work. The downside is that while the spec has "Locked until tomorrow" in red, this ends up yellow. We could do more switching on text too, but that's kind of janky...

edzkite: I agree, we shouldn't send color info from Kited, since it'd be specific to the editor. However, if the concern is code cleanliness, defining a helper function outside of here that returns a color based on the status may enable this to adhere to the spec more closely, without nesting switch cases.

edzkite (approved):

edzkite:
> I don't really have the context around these, but our terminology seems different now. Should we remove these? There's a lot of related code lying around, and I didn't want to remove everything at once.

Yeah, anything enterprise/cloud related in the plugin is old code.

lib/elements/kite-status-panel.js
tonycheang: Mainly, I'm thinking it's a little brittle to depend on the text of the message to pick a color. But I guess it probably doesn't matter too much and anything containing 'locked' probably means an error of sorts.

edzkite (approved):

PR #721, opened 2021-02-17
Add Prompt Before Installation

edzkite:
Implements updated [spec](https://kite.quip.com/VGaAAdpgwvnc/Spec-Editor-Plugin-Install-Flow)

As part of this change, the status bar will always be visible for non-supported files and while installing. 
Once Kite is installed, the status bar will be invisible if there's no active window with a filename.


tonycheang (approved):
Cool, haven't really used the emitter before, so this was insightful

PR #723, opened 2021-02-19
Fix 722

### LINKED ISSUE #722: Changing plugin settings not saved (embed codefinder decoration)
In Atom, deselecting the "embed codefinder decoration" setting is not being saved.
- Deselecting the option changes the setting for the current window, but any other windows opened will have this setting re-enabled.
- When de-selecting this setting, closing Atom, and re-opening, the setting is re-enabled. I have tried re-launching Atom with safe-mode but the issue persists, and after launching in safe mode re-closing and re-opening Atom (in standard mode) still has the issue.
- Within the same window and session (booted in normal mode), deselect this setting, disable kite, and re-enable kite - the setting is automatically re-selected.

Details:
- Kite plugin for Atom - 0.203.0
- Atom - 1.54.0

https://user-images.githubusercontent.com/61543475/108504440-c67e3280-72ad-11eb-933e-39f5f38a0c8c.mov

tonycheang:
Fixes #722 

Makes it so the setting stays what it's set to when the plugin is re-enabled

### DIFF
--- a/lib/kite.js
+++ b/lib/kite.js
@@ -199,21 +199,6 @@ const Kite = (module.exports = {
 
     // CONFIG
 
-    // Removing old configuration options
-    const toRemoveConfigs = [
-      'signatureKwargsVisible',
-      'signaturePopularPatternsVisible',
-    ];
-    const kiteCfg = atom.config.getAll('kite');
-    if (kiteCfg && kiteCfg.length > 0) {
-      const configs = Object.keys(kiteCfg[0].value);
-      configs.forEach(config => {
-        if (toRemoveConfigs.includes(config)) {
-          atom.config.unset(`kite.${config}`);
-        }
-      });
-    }
-
     this.subscriptions.add(
       atom.config.observe('kite.loggingLevel', level => {
         Logger.LEVEL = Logger.LEVELS[level.toUpperCase()];
--- a/lib/kite.js
+++ b/lib/kite.js
@@ -200,24 +200,16 @@ const Kite = (module.exports = {
     // CONFIG
 
     // Removing old configuration options
-    const kiteCfg = atom.config.getAll('kite');
-    const validKeys = [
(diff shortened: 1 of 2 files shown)


lib/kite.js
```diff
@@ -203,6 +203,7 @@ const Kite = (module.exports = {
     const kiteCfg = atom.config.getAll('kite');
     const validKeys = [
       'showWelcomeNotificationOnStartup',
+      'enableCodeFinderLineDecoration',
```
tonycheang, line 206: Quick impression is that a block list is better than an allow list so we don't have to remember to add new settings here, but I don't have the content for why this is needed (change introduced here: https://github.com/kiteco/atom-plugin/commit/28783cb0cbee20b57c72e012daef8be15b85baa0).

edzkite: It seems like it was added so that certain settings don't persist on restart, but yes I agree with your proposal.

edzkite (approved):

lib/kite.js
```diff
@@ -200,24 +200,16 @@ const Kite = (module.exports = {
...
-      'enableHoverUI',
-      'enableSnippets',
-      'maxVisibleSuggestionsAlongSignature',
-      'loggingLevel',
-      'pollingInterval',
-      'developerMode',
-      'startKiteAtStartup',
+    const toRemoveConfigs = [
+      'signatureKwargsVisible',
+      'signaturePopularPatternsVisible',
```
tonycheang, line 205: These settings weren't included in `validKeys` but I had them in my config, and they seem used in the codebase. I somewhat doubt that it's desired that these settings do not persist on restart/when re-enabling the plugin. Do you think we should just remove the "Remove old configuration options" logic altogether? @edzkite

edzkite: Yeah, I think persisting these also makes more sense to me. We can always re-add this logic if the need arises.

